Issue #4, opened 2021-06-12
Migrate log modal to Lit


balloob:
`raw_package/static/js/esphome.js` is the legacy JavaScript file. Code in there needs to be refactored into Lit/TypeScript.

This issue is to track moving the LogModal class and all implementations that rely on it (install, logs, validate).

PR #14, opened 2021-06-18
Migrate validate dialog to Lit

balloob:
Migrate the validate dialog and the foundation (log modal) to Lit.

Unblocks other migrations mentioned in #4.

### DIFF
--- a/src/components/remote-process.ts
+++ b/src/components/remote-process.ts
@@ -1,0 +1,150 @@
+import { customElement } from "lit/decorators.js";
+import { StreamError, streamLogs } from "../api";
+import { ColoredConsole } from "../util/console-color";
+import { fireEvent } from "../util/fire-event";
+
+@customElement("esphome-remote-process")
+class ESPHomeRemoteProcess extends HTMLElement {
+  public type!: "validate" | "logs" | "upload";
+  public filename!: string;
+
+  private _abortController?: AbortController;
+  private _setup = false;
+
+  public connectedCallback() {
+    if (this._setup) {
+      return;
+    }
+    this._setup = true;
+    const shadowRoot = this.attachShadow({ mode: "open" });
+
+    shadowRoot.innerHTML = `
+      <style>
+        .log {
+          height: 100%;
+          max-height: calc(100% - 56px);
+          background-color: #1c1c1c;
+          margin-top: 0;
+          margin-bottom: 0;
+          font-family: "SFMono-Regular", Consolas, "Liberation Mono", Menlo, Courier,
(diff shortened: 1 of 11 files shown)


src/components/remote-process.ts
```diff
@@ -0,0 +1,150 @@
...
+  public connectedCallback() {
+    if (this._setup) {
+      return;
+    }
+    this._setup = true;
+    const shadowRoot = this.attachShadow({ mode: "open" });
+
+    shadowRoot.innerHTML = `
+      <style>
+        .log {
```
balloob, line 23: This is copied from esphome.css. Can be removed from there once all log-based dialogs are migrated.

src/components/remote-process.ts
```diff
@@ -0,0 +1,150 @@
...
+@customElement("esphome-remote-process")
+class ESPHomeRemoteProcess extends HTMLElement {
+  public type!: "validate" | "logs" | "upload";
+  public filename!: string;
+
+  private _abortController?: AbortController;
+  private _setup = false;
+
+  public connectedCallback() {
+    if (this._setup) {
```
bramkragten, line 15: Shouldn't we restart what we aborted if we connect again?

src/install-update/index.ts
```diff
@@ -2,10 +2,10 @@ import { uploadModal } from "../legacy";
 
 const preload = () => import("./install-update-dialog");
 
-const startInstallUpdate = async (ev: Event) => {
+export const openInstallDialog = async (filename: string) => {
   preload();
   const dialog = document.createElement("esphome-install-dialog");
```
bramkragten, line 7: Do we have to create a new dialog element every time, or should we re-use it? Do we also remove the dialog when we close it?

src/validate/validate-dialog.ts
```diff
@@ -0,0 +1,86 @@
...
+
+  private _openInstall() {
+    openInstallDialog(this.filename);
+  }
+
+  private _handleProcessDone(ev: { detail: number }) {
+    this._valid = ev.detail == 0;
+  }
+
+  private async _handleClose() {
```
bramkragten, line 67:
```suggestion
  private _handleClose() {
```

bramkragten (approved):

src/install-update/index.ts
bramkragten: Yes we do.

src/components/remote-process.ts
balloob: Nah we should create a new process

PR #49, opened 2021-07-22
Underscore is a valid name character

alexiri:
Underscores are allowed according to: https://github.com/esphome/esphome/blob/dev/esphome/const.py#L9

### DIFF
--- a/src/wizard/wizard-dialog.ts
+++ b/src/wizard/wizard-dialog.ts
@@ -181,8 +181,8 @@ export class ESPHomeWizardDialog extends LitElement {
         label="Name"
         name="name"
         required
-        pattern="^[a-z0-9-]+$"
-        helper="Only use lowercase letters (a-z), numbers (0-9) or dash (-)"
+        pattern="^[a-z0-9-_]+$"
+        helper="Only use lowercase letters (a-z), numbers (0-9), dash (-) or underscore (_)"
         @input=${this._cleanNameInput}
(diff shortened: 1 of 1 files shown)


src/wizard/wizard-dialog.ts
```diff
@@ -181,8 +181,8 @@ export class ESPHomeWizardDialog extends LitElement {
         label="Name"
         name="name"
         required
-        pattern="^[a-z0-9-]+$"
-        helper="Only use lowercase letters (a-z), numbers (0-9) or dash (-)"
+        pattern="^[a-z0-9-_]+$"
+        helper="Only use lowercase letters (a-z), numbers (0-9), dash (-) or underscore (_)"
         @input=${this._cleanNameInput}
```
balloob, line 186: This method will still replace `_` with `-`.

alexiri: Sorry, missed that!

frenck:
The reason the underscore is still accepted is for legacy. As the name is used as hostname as well, the underscore is actually not valid.

For example: `http://my_test.local` isn't a valid domain. In many cases, it will gracefully work, but isn't according to specifications and can (and has) actually caused problems.

Hence, it isn't accepted for newer devices at this point.

IMHO this change should not be merged for that reason.

alexiri:
Fair enough. As it was previously valid I thought this was a regression, and not a conscious choice.

PR #63, opened 2021-08-24
Update boardlist

oxan:
For esphome/esphome#2190


balloob:
Can we create an API to allow querying this data from the backend ? That way it will always stay in sync.

oxan:
I don't see why that won't be possible (though the backend would need to be extended to know the board names and not just the IDs), but that's a lot more work and it shouldn't block updating the boardlist.

balloob (approved):

PR #76, opened 2021-09-22
Dashboard node import and render in browser

OttoWinter:
unknown

I tried to use mwc components where possible, but that wasn't possible everywhere without significant UI changes (in other cases I just didn't know CSS well enough to fix all layout issues); so `node-card.ts` ended up with a lot of copy-pasted CSS from the legacy implementation


src/api/list-entries.ts
```diff
@@ -0,0 +1,35 @@
+import { fetchApiJson, fetchApiText } from ".";
```
balloob, line 1: let's name this file `entries.ts` or maybe `nodes` ?

src/api/list-entries.ts
```diff
@@ -0,0 +1,35 @@
...
+  project_name: string;
+  project_version: string;
+}
+
+export interface ListEntriesResult {
+  configured: ConfiguredEntry[];
+  importable: ImportableEntry[];
+}
+
+export const getDashboardEntries = () =>
```
balloob, line 28: Let's call it `getEntries`. It's up to the calling method to decide how to show it.

src/components/esphome-button-menu.ts
```diff
@@ -0,0 +1,50 @@
...
+    return this._menu?.items;
+  }
+
+  public get selected() {
+    return this._menu?.selected;
+  }
+
+  protected render(): TemplateResult {
+    return html`
+      <div @click=${this._handleClick}>
```
balloob, line 22: We should use `<button>` for buttons

src/entries-list/entries-list.ts
```diff
@@ -0,0 +1,178 @@
...
+      return html``;
+    }
+    if (
+      this._entries.configured.length === 0 &&
+      this._entries.importable.length === 0
+    ) {
+      return html`
+        <div class="welcome-container">
+          <h5>Welcome to ESPHome</h5>
+          <p>It looks like you don't yet have any nodes.</p>
```
balloob, line 36: is it a node or an entry? Our API and components call it `entry`. If we still can, let's try to be consistent.

src/entries-list/entries-list.ts
```diff
@@ -0,0 +1,178 @@
...
+            (entry) => html`
+              <esphome-import-card
+                .entry=${entry}
+                @imported=${() => this._handleImported(entry)}
+              >
+              </esphome-import-card>
+            `
+          )}
+        </div>
+        <hr></hr>
```
balloob, line 67: lol

src/entries-list/entries-list.ts
```diff
@@ -0,0 +1,178 @@
...
+      `
+          : ""}
+
+        <div class="grid">
+          ${this._entries.configured.map(
+            (entry) => html` <esphome-node-card
```
balloob, line 73: If you put the HTML opening and closing tag on different lines than the string opening/closing, it formats nicer.

src/entries-list/node-card.ts
```diff
@@ -0,0 +1,266 @@
...
+import { deleteConfiguration } from "../api/configuration";
+import { fireEvent } from "../util/fire-event";
+import { openValidateDialog } from "../validate";
+import { openEditDialog } from "../legacy";
+import { openInstallDialog } from "../install-update";
+import { openLogsTargetDialog } from "../logs-target";
+
+@customElement("esphome-node-card")
+class ESPHomeNodeCard extends LitElement {
+  @property() public entry!: ConfiguredEntry;
```
balloob, line 24: node card takes an entry -> also here let's pick 1 term and stick with it.

src/entries-list/node-card.ts
```diff
@@ -0,0 +1,266 @@
...
+  @property() public onlineStatus!: boolean;
+  @property() public highlighted!: boolean;
```
balloob, line 26: Use present, not past tense

```suggestion
  @property() public highlight = false;
```

src/entries-list/node-card.ts
```diff
@@ -0,0 +1,266 @@
...
+import { fireEvent } from "../util/fire-event";
+import { openValidateDialog } from "../validate";
+import { openEditDialog } from "../legacy";
+import { openInstallDialog } from "../install-update";
+import { openLogsTargetDialog } from "../logs-target";
+
+@customElement("esphome-node-card")
+class ESPHomeNodeCard extends LitElement {
+  @property() public entry!: ConfiguredEntry;
+  @property() public onlineStatus!: boolean;
```
balloob, line 25: Default to false and you don't need to use `!` and `boolean` is derived.

src/entries-list/node-card.ts
```diff
@@ -0,0 +1,266 @@
...
+          ></mwc-button>
+          <mwc-button
+            label="Install"
+            @click=${this._handleInstall}
+          ></mwc-button>
+          <mwc-button label="Logs" @click=${this._handleLogs}></mwc-button>
+        </div>
+      </esphome-card>
+
+      <mwc-dialog>
```
balloob, line 109: Let's consider moving this to it's own component, like we do with other dialogs.

balloob (approved):
Nice!

src/entries-list/entries-list.ts
OttoWinter: Right, I guess `device` makes the most sense - we already consistently use "device" in the rest of the UI when talking about the thing running esphome.

I'll rename the APIs/components to call it `device`

src/components/esphome-button-menu.ts
OttoWinter: Hmm, then we'd have to remove all the styling from the default `<button>` - or we make the component directly use `<mwc-icon-button>` and pass in the icon (which the slot is filled with in the parent)

src/components/esphome-button-menu.ts
```diff
@@ -0,0 +1,48 @@
...
+        <slot name="trigger"></slot>
+      </div>
+      <mwc-menu .corner=${this.corner}>
+        <slot></slot>
+      </mwc-menu>
+    `;
+  }
+
+  private _handleClick(): void {
```
balloob, line 31: Whatever generates the click, (a link or a button), we should make sure to prevent any default behavior
```suggestion
  private _handleClick(ev): void {
    ev.preventDefault();
```

src/components/esphome-button-menu.ts
balloob: ok we can leave it like this. For Improv we actually add the click listener to the slot itself. https://github.com/improv-wifi/sdk-js/blob/main/src/launch-button.ts#L75-L79

balloob (approved):

src/devices/configured-device-card.ts
```diff
@@ -44,19 +44,20 @@ class ESPHomeConfiguredDeviceCard extends LitElement {
...
             ? html`
                 <div class="tooltip-container">
                   <mwc-icon class="update-available">system_update</mwc-icon>
-                  <paper-tooltip
-                    >Update Available: ${this.device.deployed_version}
-                    ${UPDATE_TO_ICON}
-                    ${this.device.current_version}</paper-tooltip
-                  >
+                  <paper-tooltip>
+                    Update Available: ${this.device.deployed_version}
```
balloob, line 58: Should there be a `<br>` here ?

OttoWinter: No, the newline is just from prettier reformatting,

It should display like this (as it was before too):

Update Available: {old_ver} :arrow_right: {new_ver}

PR #88, opened 2021-10-08
Fix web_server break device card

OttoWinter:
Typo from a refactor, previously the property was called `entry`, but I changed it to `device` for consistency. Forgot to rename it here


poldim:
Still seeing this issue on the latest dev as of right now.  This is what my dashboard shows:

PR #100, opened 2021-10-14
Hide online status if online

balloob:
Hide the online status text if the device is online. Only show the text if the device is unknown or offline.

We initially had removed all text, but reinstated that because colorblind people could not distinguish the top border of the card. The "online" case is the positive case and should be assumed. We now only show "Unknown" or "Offline" (the error states).

image


jesserockz:
But then this makes the cards have different heights? Would it not be better to have them all the same?

balloob:
We already have different heights too if we have descriptions

PR #104, opened 2021-10-18
Card tweaks

### LINKED ISSUE #103: Different card height / jumping cards in 2021.10.0b4
In ESPHome 2021.10.0b4 the online state is hidden. Because of that, the height of the cards are different:
image
While I like the more compact view - it causes all cards to jump up/down when the status of one of the upper devices changes and also doesn't look clean.

Suggestion:
Use the icon of HA binary sensor device class connectivity to show if a device is offline (with tooltip & alt text "OFFLINE"):
image

balloob:
Update cards:
 - Move overflow menu to action bar
 - Change offline text to icon and move to action bar
 - Remove unknown online status, checked with Jesse and it cannot happen anymore
 - Move install to overflow menu
 - Move open in web ui to action bar ~~and disable if device offline~~
image

Fixes #103

### DIFF
--- a/raw_package/static/css/esphome.css
+++ b/raw_package/static/css/esphome.css
@@ -170,13 +170,6 @@ nav {
   color: black;
 }
 
-main .container {
-  margin-top: 20px;
-  margin-bottom: 20px;
-  width: 90%;
-  max-width: 1920px;
-}
-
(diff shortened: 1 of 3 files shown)


balloob:
Design implemented 👍

PR #125, opened 2021-10-20
clean up card status

balloob:
- Move all card status logic to `<esphome-card>`.
- Move 'highlight' to be shown as a status for 10 seconds
- Automatically highlight any new device added after first render (import/wizard/new file on disk)
- automate the status color for a second to draw attention
- Move status text slightly up

https://user-images.githubusercontent.com/1444314/138026978-59285c04-fb92-4bb0-a775-4f9456087166.mp4

### DIFF
--- a/src/devices/configured-device-card.ts
+++ b/src/devices/configured-device-card.ts
@@ -8,7 +8,6 @@ import "@material/mwc-button";
 import "@material/mwc-icon-button";
 import "../components/esphome-button-menu";
 import "../components/esphome-card";
-import "./delete-device-dialog";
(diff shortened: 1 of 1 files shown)


balloob:
I am only going to respond to feedback related to the changes in this PR. If you have any other comments, feel free to open an issue.

- Moving the label down and pushing long headers to a second line means that cards will shift around because all of a sudden the text is broken down.
- I'll changed the word "highlight" to "new" 
- I'll play with the popping of the card

For comparison, this is the current highlight feature in ESPHome dashboard:

https://user-images.githubusercontent.com/1444314/138116198-ee998443-fd4b-435e-9228-424673e5124c.mp4

PR #127, opened 2021-10-21
Fix login.template.html css url

OttoWinter:
Fixes https://github.com/esphome/issues/issues/2572


balloob:
oops. So much for my cache busting techniques.

PR #139, opened 2021-11-08
Also allow dot secrets dot yaml

cvwillegen:
Related PR in esphome/esphome#2649. This allows for specifying secrets in a file called .secrets.yaml. This PR adds the possibility to edit this file from the Dashboard.

### DIFF
--- a/src/components/esphome-header-menu.ts
+++ b/src/components/esphome-header-menu.ts
@@ -26,6 +26,11 @@ export class ESPHomeHeaderMenu extends LitElement {
         ></mwc-button>
         <mwc-button
           icon="lock"
+          label=".Secrets"
+          @click=${this._handleEditDotSecrets}
+        ></mwc-button>
+        <mwc-button
+          icon="lock"
           label="Secrets"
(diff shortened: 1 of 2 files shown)


src/components/esphome-header-menu.ts
```diff
@@ -46,6 +51,7 @@ export class ESPHomeHeaderMenu extends LitElement {
       >
         <mwc-icon-button slot="trigger" icon="more_vert"></mwc-icon-button>
         <mwc-list-item>Update All</mwc-list-item>
+        <mwc-list-item>.Secrets Editor</mwc-list-item>
```
balloob, line 54: We shouldn't add this. There is one way of editing secrets in the frontend and that's the other menu item, regardless of where the files are stored in the backend.

balloob (changes requested):

cvwillegen:
I will follow the comment in esphome/esphome#2649 from @balloob and skip (or warn) at files called `secrets.yaml` and `secrets.yml`

PR #140, opened 2021-11-10
Allow webSerial via trusted http hosts

jesserockz:


src/const.ts
```diff
@@ -2,8 +2,6 @@ import { CHIP_FAMILY_ESP32, CHIP_FAMILY_ESP8266 } from "esp-web-flasher";
 import { svg } from "lit";
 
 export const supportsWebSerial = "serial" in navigator;
-export const allowsWebSerial =
```
balloob, line 5: This is not the right approach. Check for secure context instead like this https://github.com/esphome/esp-web-tools/blob/main/src/install-button.ts#L4-L6

balloob (approved):

PR #186, opened 2022-01-11
Allow configuring Wi-Fi

balloob:
Add menu option to configure wifi. We already automatically opened this after making a device adoptable.

image


jesserockz (approved):
:rocket:

PR #285, opened 2022-08-20
migrate editor add hover help and completions

### LINKED ISSUE #5: Migrate editor to Lit
`raw_package/static/js/esphome.js` is the legacy JavaScript file. Code in there needs to be refactored into Lit/TypeScript.

This issue is to track moving the editor.

glmnet:
Fixes #5

Todo:

- [ ] UI for asynchronous loading feedback of yaml files 
- [ ] test packaging inside ESPHome (manifest.in file updates?)


src/editor/esphome-editor.ts
```diff
@@ -0,0 +1,289 @@
...
+// WebSocket URL Helper
+const loc = window.location;
+const wsLoc = new URL("./", `${loc.protocol}//${loc.host}${loc.pathname}`);
+wsLoc.protocol = "ws:";
+if (loc.protocol === "https:") {
+  wsLoc.protocol = "wss:";
+}
+const wsUrl = wsLoc.href;
+
+let editorActiveFilename: string | null = null;
```
balloob, line 18: Why are these globals?

src/editor/esphome-editor.ts
```diff
@@ -0,0 +1,289 @@
...
+let editorActiveSecrets = false;
+let editorActiveWebSocket: WebSocket | null = null;
+let editorValidationScheduled = false;
+let editorValidationRunning = false;
+
+@customElement("esphome-editor")
```
balloob, line 24: Have you considered using https://github.com/vanillawc/wc-monaco-editor

src/editor/esphome-editor.ts
```diff
@@ -0,0 +1,289 @@
...
+    });
+    // window
+    //   .matchMedia("(prefers-color-scheme: dark)")
+    //   .addEventListener("change", () => {
+    //     monaco.editor.setTheme(this.getTheme());
+    //   });
+    const filename = this.configuration;
+    editorActiveFilename = filename;
+    const isSecrets = filename === "secrets.yaml" || filename === "secrets.yml";
+    const filenameField = document.querySelector(
```
balloob, line 108: If we need to query things, we should query local using `this.shadowRoot!.querySelector(...)`

src/editor/editor-dialog.ts
```diff
@@ -0,0 +1,118 @@
...
+  private handleInstall() {
+    this._saveFile();
+    this._handleClose();
+    openInstallChooseDialog(this.fileName);
+  }
+
+  private _saveFile() {
+    const code = this.editorRef.value!.getValue();
+
+    fetch(`./edit?configuration=${this.fileName}`, {
```
balloob, line 89: All data handling should be moved to https://github.com/esphome/dashboard/tree/main/src/api

src/editor/esphome-editor.ts
```diff
@@ -0,0 +1,289 @@
...
+    });
+
+    editorActiveWebSocket.addEventListener("close", () => {
+      editorActiveWebSocket = null;
+      setTimeout(this.startAceWebsocket, 5000);
+    });
+  }
+}
+
+const debounce = (func: Function, wait: number) => {
```
balloob, line 244: debouce should be moved to util folder.

src/editor/esphome-editor.ts
glmnet: no, in fact I tried other alternatives without much success. This project you point is a bit unmaintained and also I could not understand how to get access to the monaco api with this (may be I just don't get it) 
Do you really want to focus on this library or is this just a shallow question?
also I'm not sure why would we use that one instead of what I've done now

balloob: The main reason would be maintenance. I didn't look into that web component specifically. More a generic question why not re-use something that will figure out all the CSS etc for us.

glmnet:
I added some progress, I had a lot of refactoring going on to get this in sync with vscode add on, so "main" files are in sync and there are some "adaptors", even thou they both now use monaco editor the apis are slightly different so had to account that.

> Have you considered using https://github.com/vanillawc/wc-monaco-editor

I have nothing against using/changing how monaco is included, but I couldn't figure out how to use the component.

I've removed what I could of legacy and ace, not sure if jquery is still used.

the schemas (json files generated from esphome/docs) are bundled, they could come from the cloud but may be this way is better, it's a bit of egg and chicken when releasing, schemas should be updated during beta stage, and should be ok, seems to be about 2.2mb before zip

never did anything regarding the manifest.in thing which I guess will need to be updated to include the json schema files (nor do I know how to fully test a package) i could use some help there when needed

Misiu:
@glmnet I think I was able to remove jQuery - https://github.com/esphome/dashboard/pull/286
I'd be grateful if you could take a look at my PR and give it a try.

Misiu:
I think you can remove jQuery. It was only used in legacy.js + single place in login.template, [this one line](https://github.com/esphome/dashboard/pull/286/files#diff-6c294fea5135098609ae91d0d227ae8172e083bb890d6a4ea553bef4d6019eafL81) this will fix that file

glmnet:
BTW in case it wasn't noticed the completions/hover code is (manually) [in sync](https://github.com/esphome/esphome-vscode/tree/main/server/src) with vscode. [There are several test cases ](https://github.com/esphome/esphome-vscode/tree/main/server/test) on completions and a few other stuff

rollup.config.js
```diff
@@ -21,17 +28,46 @@ const config = {
...
+            fs.ensureDirSync(distFontsPath);
+            const targetFontPath = path.join(
+              "esphome_dashboard/static/fonts/",
+              asset.pathname
+            );
+            fs.copySync(asset.absolutePath, targetFontPath);
+            const relativePath = path.relative(process.cwd(), targetFontPath);
+            console.log(relativePath);
+            const publicPath = "/";
+            return "/static/fonts/" + asset.pathname; // `${publicPath}${relativePath}`;
```
balloob, line 47:
```suggestion
            const publicPath = "/";
            return "/static/fonts/" + asset.pathname;
```

rollup.config.js
```diff
@@ -21,17 +28,46 @@ const config = {
...
+          },
+        }),
+      ],
+    }), copy({
+      targets: [
+        { src: 'schema/*.json', dest: 'esphome_dashboard/static/schema' },
```
balloob, line 53: Can we download this once, and then reuse that?

1. Add `schemas` to `.gitignore`
1. When rollup config is loaded and `schemas` doesn't exist, download it
2. keep this copy command

balloob: And maybe a script to sync schemas that re-downloads it.

PR #286, opened 2022-09-05
remove jQuery

Misiu:
Materialize 1.0 doesn't require jQuery as a dependency - https://medium.com/@materializecss/materialize-to-1-0-and-beyond-e0233b8ac15

I replaced `$(document).ready(function () {` with `document.addEventListener("DOMContentLoaded",()=>{});` and it works well after quick testing (I have ESPHome running on Docker).

I wasn't able to test changes in `login.template.html` but the change there is very simple.


glmnet:
This will require the legacy to be dropped so we need the editor migrated or something else done before. 

Thanks!

Misiu:
@glmnet that's the plan - to drop legacy code. 
This PR is the first step, it only drops jQuery, but the functionality is unchanged.
As mentioned I replaced document.ready in legacy.js, all other functions in that file don't use jQuery, so I left them untouched.
Materialize js (which is used for toast messages) should be the next step.
I think that Monaco change should go first and then all the rest of the legacy code could be removed.

I'm not sure what browsers and in which versions are supported by the dashboard, that's something that should be added to the readme.

glmnet:
My bad. I though you deleted the whole legacy.js already

Misiu:
> My bad. I though you deleted the whole legacy.js already

That's the next step :)
legacy.js has editor (ace) code, so after your PR gets merged we can remove the remaining parts.

PR #295, opened 2022-09-20
Remove materialize.js and use mwc-snackbar instead

Misiu:
materialize.min.js was just used to show toasts in editor-dialog.
Now we use mwc-snackbar (as suggested in https://github.com/esphome/dashboard/issues/5), so materialize.min.js can be finally removed.

I tested index.template as I was unable to test login.template.

Feedback welcome.


balloob (approved):
Nice! 🎉 

Pushed a little cleanup.

PR #307, opened 2022-09-22
Add lit-analyzer, prettier and add missing closing tag

Misiu:
I noticed that `device-card.ts` had a missing `div` tag, this adds it, hopefully in the right place.
Btw how to test web.esphome on dev machine?

I've also added additional tasks to package.json, they will allow formatting code and finding problems in lit (currently 20 problems in 11 files).

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,7 +7,11 @@
   "author": "ESPHome maintainers",
   "license": "Apache-2.0",
   "scripts": {
-    "prepublishOnly": "script/build"
+    "prepublishOnly": "script/build",
+    "lint:prettier": "prettier \"src/**/*.{js,ts,json,css,md}\" --check",
+    "format:prettier": "prettier \"src/**/*.{js,ts,json,css,md}\" --write",
+    "lint:types": "tsc",
+    "lint:lit": "lit-analyzer \"src/**/*.ts\""
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^22.0.2",
--- a/package.json
+++ b/package.json
@@ -45,6 +45,7 @@
     "esp-web-flasher": "^5.1.4",
     "improv-wifi-serial-sdk": "^2.2.2",
     "lit": "^2.3.1",
+    "lit-analyzer": "^1.2.1",
     "tslib": "^2.4.0",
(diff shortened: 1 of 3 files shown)


balloob:
Test web with running `./web.esphome.io/script/develop_web`

web.esphome.io/src/dashboard/device-card.ts
```diff
@@ -53,6 +47,7 @@ class EWDeviceCard extends LitElement {
             <mwc-list-item>Configure Wi-Fi</mwc-list-item>
             <mwc-list-item>Disconnect</mwc-list-item>
           </esphome-button-menu>
+        </div>
```
balloob, line 50: Please split this out into a new PR.

package.json
```diff
@@ -7,7 +7,11 @@
   "author": "ESPHome maintainers",
   "license": "Apache-2.0",
   "scripts": {
-    "prepublishOnly": "script/build"
+    "prepublishOnly": "script/build",
+    "lint:prettier": "prettier \"src/**/*.{js,ts,json,css,md}\" --check",
+    "format:prettier": "prettier \"src/**/*.{js,ts,json,css,md}\" --write",
+    "lint:types": "tsc",
+    "lint:lit": "lit-analyzer \"src/**/*.ts\""
```
balloob, line 14: I don't see much need for these commands?

web.esphome.io/src/dashboard/device-card.ts
Misiu: Done in https://github.com/esphome/dashboard/pull/309

package.json
Misiu: My idea was to mimic actions that we have in HA Frontend and maybe add them in pre-commit. Especially prettier.

balloob: In an ideal world, we have a pre-commit, and then run pre-commit in check mode in CI. Pre-commit is Python based, not sure if there is a JS variant or that we require people to have Python? (not the end of the world)

But if we do pre-commit, would we still need the tasks here? And we shouldn't add tasks unless we have pre-commit in place

balloob: yeah, let's mimic HA frontend 👍

PR #311, opened 2022-09-23
fix dim font

glmnet:
Fixes thin font in iOS, I expect this to fix it in other OSs too.

### DIFF
--- a/src/editor/esphome-editor.ts
+++ b/src/editor/esphome-editor.ts
@@ -62,7 +62,7 @@ export class ESPHomeEditor extends LitElement {
       },
       tabSize: 2,
       fontFamily:
-        "ui-monospace,SFMono-Regular,Menlo,Monaco,Consolas,Liberation Mono,Courier New,monospace",
+        'ui-monospace,SFMono-Regular,Menlo,Monaco,Consolas,"Liberation Mono","Courier New",monospace',
     });
 
     const filename = this.configuration;
--- a/src/editor/esphome-editor.ts
+++ b/src/editor/esphome-editor.ts
@@ -62,7 +62,7 @@ export class ESPHomeEditor extends LitElement {
       },
       tabSize: 2,
       fontFamily:
-        "SFMono-Regular, Consolas, Liberation Mono, Menlo, Courier, monospace",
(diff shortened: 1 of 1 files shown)


src/editor/esphome-editor.ts
```diff
@@ -61,6 +61,8 @@ export class ESPHomeEditor extends LitElement {
         enabled: false,
       },
       tabSize: 2,
+      fontFamily:
+        "SFMono-Regular, Consolas, Liberation Mono, Menlo, Courier, monospace",
```
balloob, line 65: Add Roboto Mono to it.

balloob: Try this https://qwtel.com/posts/software/the-monospaced-system-ui-css-font-stack/#the-monospaced-system-ui-css-font-stack

balloob: Let's just copy what Tailwind uses

https://github.com/tailwindlabs/tailwindcss/blob/master/src/css/preflight.css#L106

balloob: Also don't wrap the whole value in quotes, that is not supported

glmnet: too many quotes there

glmnet: > Also don't wrap the whole value in quotes, that is not supported

this is not a css style rule, it's a js object

src/editor/esphome-editor.ts
```diff
@@ -61,6 +61,8 @@ export class ESPHomeEditor extends LitElement {
         enabled: false,
       },
       tabSize: 2,
+      fontFamily:
+        "ui-monospace,SFMono-Regular,Menlo,Monaco,Consolas,Liberation Mono,Courier New,monospace",
```
balloob, line 65: I think that liberation mono still needs to be wrapped in quotes to be a valid font family value, no ?
```suggestion
        "ui-monospace,SFMono-Regular,Menlo,Monaco,Consolas,\"Liberation Mono\",\"Courier New\",monospace",
```

balloob (approved):

src/editor/esphome-editor.ts
glmnet: you're right, spaces needs to be escaped

PR #312, opened 2022-09-23
Fullscreen-edit

### LINKED ISSUE #304: Editor scroll on mobile
Please add the scroll ability to the config editor on mobile devices.
I understand that there is horizontal and vertical scrollbars, but using them is uncomfortable.

glmnet:
this is an attempt to get it to full screen edit mode.

I guess may be a router will be a better choice, back button will send you out of the editor, however tornado is picking up routes and, e.g. /edit is handled as an action to do something with the files, so we would have to move those to api/ or find another solution. I'm not proposing to leave SPA (can do client routing) but I'm ok if we have another template for the editor, that might even tidy up stuff.

~Scrollbar are messed up in this commit, I really cannot follow how to do these layouts, with floating bars at top and bottom. Also resizing the window just get stuff worse. Even this mess, I still prefer this to what we have before.~

This fixes the issue of the editor scroll not working on mobile. This is a bug in monaco-edior which I did not investage too much but tried to avoid it, the problem is that it just doesn't work when is in a shadowroot so I have used the `createRenderRoot() { return this; }` code found somewhere to avoid it and got it working, I couldn't do this inside the popup.

fixes #304


src/editor/esphome-editor.ts
```diff
@@ -18,56 +24,131 @@ if (loc.protocol === "https:") {
...
-  render() {
+  protected render() {
+    const isSecrets =
+      this.fileName === "secrets.yaml" || this.fileName === "secrets.yml";
+
     return html`
       <style>
-        ${editorStyles}
-        ${ESPHomeEditor.styles}
+        editor-container {
```
balloob, line 48: Why is this inline and not using `static styles = css`

glmnet: it  stops working when the `createRenderRoot() { return this; } ` thing is added

src/index.ts
```diff
@@ -1,3 +1,52 @@
 import "./devices/devices-list";
 import "./components/esphome-header-menu";
 import "./components/esphome-fab";
+import "./editor/esphome-editor";
```
balloob, line 4: We should not load the editor in the entrypoint. This means it will load the full editor just to show the cards! Instead, add `import("./editor/esphome-editor");` to firstUpdated()

src/index.ts
```diff
@@ -1,3 +1,52 @@
...
+import { LitElement, html } from "lit";
+import { customElement, state } from "lit/decorators.js";
+import { editCallback } from "./editor";
+
+@customElement("esphome-main")
+class ESPHomeMainView extends LitElement {
+  @state() private editing?: string;
+  constructor() {
+    super();
+    editCallback.callback = (configuration: string) => {
```
balloob, line 14: What is this? And do we set a global?

src/index.ts
```diff
@@ -1,3 +1,52 @@
...
+      this.editing = configuration;
+    };
+  }
+
+  protected render() {
+    if (this.editing) {
+      return html`<style>
```
balloob, line 21: We should use `static styles = css`

src/index.ts
balloob: You already listen to close event on the editor, isn't that the same?

src/index.ts
glmnet: ok, thanks for the hint

src/editor/esphome-editor.ts
```diff
@@ -18,56 +24,131 @@ if (loc.protocol === "https:") {
...
+  }
+
   firstUpdated() {
     // @ts-ignore
     self.MonacoEnvironment = {
       getWorkerUrl: function (moduleId: string, label: string) {
         return "./static/js/esphome/monaco-editor/esm/vs/editor/editor.worker.js";
       },
     };
+    //this.container.style.height = "70vh";
```
glmnet, line 135:
```suggestion
```

src/index.ts
glmnet: again, see line 37: it does not work when skipping the shadowRoot with 
```
createRenderRoot() {
  return this;
}

File: src/index.ts (hunk 3 above)
Comment (glmnet) [reply src/index.ts:14]: if you have a better approach to show up the editor please give more info on how to do that

Comment (balloob) [reply src/index.ts:14]: fire an event. That's how we do it in the HA frontend. Events bubble up through the DOM, so you need to make sure you're listening on an element that gets all the events from everywhere, like `document.body`

File: src/index.ts (hunk 4 above)
Comment (balloob) [reply src/index.ts:21]: Ah that's correct.

File: src/editor/esphome-editor.ts [hunk 6]
```diff
@@ -191,16 +286,23 @@ export class ESPHomeEditor extends LitElement {
...
+  }
+  connectedCallback() {
+    super.connectedCallback();
+    window.addEventListener("resize", this._handleResize);
+  }
+  disconnectedCallback() {
+    window.removeEventListener("resize", this._handleResize);
+    super.disconnectedCallback();
+  }
+  _handleResize() {
```
balloob, line 303: This needs to be a bound method or else `this` doesn't refer to this element
```suggestion
  _handleResize = () => {
```

src/index.ts
```diff
@@ -1,3 +1,90 @@
 import "./devices/devices-list";
 import "./components/esphome-header-menu";
 import "./components/esphome-fab";
+import { LitElement, html, PropertyValues } from "lit";
+import { customElement, state } from "lit/decorators.js";
+
+const esphome_globals: {
+  version: string;
+  docs_link: string;
+  logoutUrl: string;
```
balloob, line 10:
```suggestion
  logoutUrl?: string;
```

balloob (approved):

PR #313, opened 2022-09-23
create extensions.json from ha repo

glmnet:
are all of these recommended here too?

### DIFF
--- a/.vscode/extensions.json
+++ b/.vscode/extensions.json
@@ -1,0 +1,8 @@
+{
+  "recommendations": [
+    "dbaeumer.vscode-eslint",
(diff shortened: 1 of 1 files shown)


Misiu:
This file is added to gitignore in HA Frontend https://github.com/home-assistant/frontend/blob/dev/.gitignore#L34, we should do the same here.

balloob:
it's added to ignore after the file was created I think. I am not 100% sure the reasoning.

glmnet:
Ohh I created directly in GitHub so maybe an update to .gitignore is required

Misiu:
I think those four are a good strart. We can always add/edit this in the future.

PR #322, opened 2022-09-29
Remove materialize.min.css

Misiu:
I've removed the reference to `materialize.min.css` in both layouts (btw why do we need two base layouts??)
I've also added a variable for footer color.
In the next PR I'd like to cleanup CSS, extract colors into variables and try to add dark theme (based on colors from HA Frontend)
Closes https://github.com/esphome/dashboard/issues/174


glmnet:
If you mean the new base layout is because of the other app ESPHome-web.

Misiu:
I didn't check that one with my changes.
I'm on windows and it's hard to run dev env with scripts.
Ideally, this should be done via npm run, but I can work on that in the future.
@glmnet could you please check if my changes don't affect ESPHome-web?

raw_package/base.template.html
```diff
@@ -20,18 +20,13 @@
...
     <link
       rel="stylesheet"
       href="{{ get_static_file_url('css/esphome-2.css') }}"
     />
   </head>
 
   <body>
-    {% block main %}{% end %}
-    {% block footer %}{% end %}
+    {% block main %}{% end %} {% block footer %}{% end %}
```
balloob, line 30: Please revert this change.

balloob:
We have 2 base layouts because ESPHome and ESPHome Web use different template engines which prevents us from using multiple inheritance. Ideally we would migrate both to the same.

raw_package/base.template.html
Misiu: prettier formatted this. Give me a sec and I revert it.

Misiu: done

PR #323, opened 2022-09-30
Add icons to device card

Misiu:
`esphome-svg-icon` is a copy of svg icon from HA Frontend, so that part was easy :)
Below are icons propositions:
image

I'm open to suggestions for more suitable icons

### DIFF
--- a/src/components/esphome-svg-icon.ts
+++ b/src/components/esphome-svg-icon.ts
@@ -1,0 +1,51 @@
+import { css, CSSResultGroup, LitElement, svg, SVGTemplateResult } from "lit";
+import { customElement, property } from "lit/decorators.js";
+
+@customElement("esphome-svg-icon")
+export class ESPHomeSvgIcon extends LitElement {
+  @property() public path?: string;
+
+  @property() public viewBox?: string;
+
+  protected render(): SVGTemplateResult {
+    console.log(this.path);
+    return svg`
+    <svg
+      viewBox=${this.viewBox || "0 0 24 24"}
+      preserveAspectRatio="xMidYMid meet"
+      focusable="false"
+      role="img" 
+      aria-hidden="true"
+    >
+      <g>
+      ${this.path ? svg`<path d=${this.path}></path>` : ""}
+      </g>
+    </svg>`;
+  }
+
+  static get styles(): CSSResultGroup {
+    return css`
+      :host {
+        display: var(--ha-icon-display, inline-flex);
+        align-items: center;
+        justify-content: center;
+        position: relative;
+        vertical-align: middle;
+        fill: currentcolor;
+        width: var(--mdc-icon-size, 24px);
+        height: var(--mdc-icon-size, 24px);
+      }
+      svg {
+        width: 100%;
+        height: 100%;
+        pointer-events: none;
+        display: block;
+      }
+    `;
+  }
+}
+declare global {
+  interface HTMLElementTagNameMap {
+    "esphome-svg-icon": ESPHomeSvgIcon;
+  }
+}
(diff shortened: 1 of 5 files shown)


src/components/esphome-svg-icon.ts
```diff
@@ -0,0 +1,51 @@
...
+import { customElement, property } from "lit/decorators.js";
+
+@customElement("esphome-svg-icon")
+export class ESPHomeSvgIcon extends LitElement {
+  @property() public path?: string;
+
+  @property() public viewBox?: string;
+
+  protected render(): SVGTemplateResult {
+    console.log(this.path);
```
balloob, line 11:
```suggestion
```

src/devices/configured-device-card.ts
```diff
@@ -130,15 +140,57 @@ class ESPHomeConfiguredDeviceCard extends LitElement {
...
+              ></esphome-svg-icon
+            ></mwc-list-item>
+            <li divider role="separator"></li>
+            <mwc-list-item class="warning" graphic="icon"
+              >Delete<esphome-svg-icon
+                class="warning"
+                slot="graphic"
+                .path=${mdiDelete}
+              ></esphome-svg-icon
+            ></mwc-list-item>
```
balloob, line 186: Can we render this a bit nicer. Like this:
```suggestion
            <mwc-list-item class="warning" graphic="icon">
             <esphome-svg-icon
                class="warning"
                slot="graphic"
                .path=${mdiDelete}
              ></esphome-svg-icon>
              Delete
            </mwc-list-item>
```

balloob (approved):

src/components/esphome-svg-icon.ts
Misiu: oops

balloob:
The icon colors are a bit off. It makes it look like the option is disabled. In HA we use secondary-text color, can you adopt that?

src/devices/configured-device-card.ts
Misiu: prettier formatted the code this way (I have FOrmat on Save in VS Code enabled). I think the checks won't pass if I change formatting, but I can try that.

balloob:
Suggestion for validate:
image

balloob:
Suggestion for install
image

(we should avoid using outline icons unless we do all outline icons)

glmnet:
> Suggestion for install image
> 
> (we should avoid using outline icons unless we do all outline icons)

any thinking for a down arrow? It feels like retrieving stuff not pushing stuff

May be if you are pushing firmware from cloud (esphome.web) to a node locally that makes more sense but from my desktop (or even local running HA instance) to my iot thing in the ceiling not so much

bramkragten:
CleanShot 2022-09-30 at 15 57 39

for clean build files maybe?

bramkragten:
> > Suggestion for install image
> > (we should avoid using outline icons unless we do all outline icons)
> 
> any thinking for a down arrow? It feels like retrieving stuff not pushing stuff
> 
> May be if you are pushing firmware from cloud (esphome.web) to a node locally that makes more sense but from my desktop (or even local running HA instance) to my iot thing in the ceiling not so much

Agree, upload is better:
CleanShot 2022-09-30 at 16 00 12

Misiu:
I've updated the icons.

balloob:
I thought down arrow because it's going into the device connected to the network. But up works too.

balloob:
@Misiu don't forget to push your commit 👍

Misiu:
@balloob sorry for that :) I've pushed the changes.
Not.sure about the menu items and icon colors, I didn't change them. I can tweak them in future PR (I want to align them with HA and try to add a dark theme)

PR #327, opened 2022-10-04
Update style of status text

bramkragten:
I think this looks nicer and fits the overall design better.

Old:
image
New:
image


balloob:
The status text was so high so long titles wouldn't overlap with the status text.

PR #339, opened 2022-10-17
Add list-view

masterwendu:
Hello,

Add additional view option list.
Can be toggled via a button in the header menu.
View mode is saved in the localStorage of the browser.

This list-view should help to find devices faster in the list.

See preview:
list-view-preview

This is my first work with lit. I am very open to feedback :smiley:

### DIFF
--- a/src/esphome-main.ts
+++ b/src/esphome-main.ts
@@ -47,7 +47,7 @@ class ESPHomeMainView extends LitElement {
           alt="ESPHome Logo"
         />
         <div class="flex"></div>
-        <esphome-header-menu .logoutUrl=${this.logoutUrl} .toggleViewMode=${this._toggleViewMode.bind(this)} .viewMode=${this.viewMode}></esphome-header-menu>
+        <esphome-header-menu .logoutUrl=${this.logoutUrl} @toggle-view-mode=${this._toggleViewMode} .viewMode=${this.viewMode}></esphome-header-menu>
       </div>
       <esphome-fab></esphome-fab>
       <footer class="page-footer grey darken-4">
--- a/src/esphome-main.ts
+++ b/src/esphome-main.ts
@@ -3,6 +3,9 @@ import "./components/esphome-header-menu";
 import "./components/esphome-fab";
 import { LitElement, html, PropertyValues } from "lit";
 import { customElement, property, state } from "lit/decorators.js";
+import ViewMode from "./devices/viewMode"
+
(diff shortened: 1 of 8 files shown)


src/esphome-main.ts
```diff
@@ -32,15 +39,15 @@ class ESPHomeMainView extends LitElement {
...
+        <esphome-devices-list .viewMode=${this._isWide ? this.viewMode : ViewMode.Module}></esphome-devices-list>
       </main>
       <div class="esphome-header">
         <img
           src="https://esphome.io/_static/logo-text.svg"
           alt="ESPHome Logo"
         />
         <div class="flex"></div>
-        <esphome-header-menu .logoutUrl=${this.logoutUrl}></esphome-header-menu>
+        <esphome-header-menu .logoutUrl=${this.logoutUrl} .toggleViewMode=${this._toggleViewMode.bind(this)} .viewMode=${this.viewMode}></esphome-header-menu>
```
Misiu, line 50: maybe instead of passing a function consider dispatching a custom event and listening to that event.
Ref: https://github.com/esphome/dashboard/blob/2a59a7875908a5bda9e19b033e947a074b85e00e/src/components/remote-process.ts#L47

masterwendu: Thanks for your feedback :tada:  I updated the code.

balloob:
So I don't think that is an improvement. For starters it doesn't work on mobile.

masterwendu:
> So I don't think that is an improvement. For starters it doesn't work on mobile. 

For me the improvement is that I have a better overview over all my devices and when I use a consistent naming I can find them quick. 

Next step would be to integrate a search functionality. This would be also an improvement for mobile. 

Can you explain what you mean with starters?

Mobile I didn't implemented because there is already a list of modules (cards). 
It would be possible when we add the buttons to the card menu or exchange the buttons with icons.

balloob:
- The change view button is put next to action buttons in header
- The cards are mashed into rows, causing the rows not to be nicely styled
- Doesn't work on mobile
- Trying to make the cards with hacked up CSS work as a row, including on small screens is going to unmaintainable 

If we want to explore a list UI, we need to start with a mockup. If we then want to proceed, it needs it's own set of components. Not reusing the card component.

However, the problem now is that this doubles maintenance and we need to do every change to devices twice. Is the squeeze worth the juice?

masterwendu:
Ok. I understand. Thanks for your detailed feedback. Maybe next time.

PR #343, opened 2022-10-19
add --legacy-peer-deps in readme

glmnet:


balloob:
Do we know why it is needed?

glmnet:
Well, added when we added Monaco editor, otherwise no npm install won't work. 

Is something about how packages (other packages) references their dependencies, e.g Monaco-roll up-plug-in references Monaco, it used to work but then newer npm version doesn't like it anymore.

PR #346, opened 2022-10-21
Add UI support for Seeed Studio XIAO ESP32C3

lakshanthad:
Hello,

Please help to merge the Seeed Studio XIAO ESP32C3 board support to ESPHome Dashboard.

Here is board support on PlatformIO for reference:
https://github.com/platformio/platform-espressif32/blob/develop/boards/seeed_xiao_esp32c3.json

Thank you.

### DIFF
--- a/src/wizard/boards.ts
+++ b/src/wizard/boards.ts
@@ -21,6 +21,18 @@ export const boardSelectOptions = html`
     <option value="thing">Sparkfun ESP8266 Thing</option>
     <option value="thingdev">Sparkfun ESP8266 Thing - Dev Board</option>
   </optgroup>
+  <optgroup label="ESP32C3">
+    <option value="adafruit_qtpy_esp32c3">Adafruit QT Py ESP32-C3</option>
+    <option value="airm2m_core_esp32c3">AirM2M CORE ESP32C3</option>
+    <option value="wifiduino32c3">Blinker WiFiduinoV2 (ESP32-C3)</option>
+    <option value="dfrobot_beetle_esp32c3">DFRobot Beetle ESP32-C3</option>
+    <option value="deneyapkartg">Deneyap Kart G</option>
+    <option value="esp32-c3-devkitm-1">Espressif ESP32-C3-DevKitM-1</option>
+    <option value="seeed_xiao_esp32c3">Seeed Studio XIAO ESP32C3</option>
+    <option value="bee_motion_mini">Smart Bee Motion Mini</option>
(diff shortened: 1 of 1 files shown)


src/wizard/boards.ts
```diff
@@ -8,6 +8,7 @@ export const boardSelectOptions = html`
     <option value="lolin_d32_pro">Wemos Lolin D32 Pro</option>
     <option value="featheresp32">Adafruit ESP32 Feather</option>
     <option value="m5stack-core-esp32">M5Stack Core ESP32</option>
+    <option value="seeed_xiao_esp32c3">Seeed Studio XIAO ESP32C3</option>
```
balloob, line 11: Can you add it to the "Other ESP32s" section.

lakshanthad: Can I add to the start of "Other ESP32s" section?

lakshanthad: Also, will it use the correct "variant" as "esp32c3"?

balloob: > Can I add to the start of "Other ESP32s" section?

We're not a marketing tool. Let's sort this alphabetically.

Maybe we should break out the C3.

lakshanthad: > Maybe we should break out the C3.

Do you mean to have a new section for ESP32C3?

It seems there is a number of ESP32C3 boards available already.

balloob: Yeah that would probably make more sense.

lakshanthad: Hi. Sorry for late response.

I have added all the ESP32C3 boards under "ESP32C3" section

src/wizard/boards.ts
```diff
@@ -21,6 +21,18 @@ export const boardSelectOptions = html`
     <option value="thing">Sparkfun ESP8266 Thing</option>
     <option value="thingdev">Sparkfun ESP8266 Thing - Dev Board</option>
   </optgroup>
+  <optgroup label="ESP32C3">
```
balloob, line 24:
```suggestion
  <optgroup label="ESP32-C3">
```

balloob (approved):

balloob:
I was talking to @jesserockz and he mentioned that we use [`v3.5.0`](https://github.com/esphome/esphome/blob/dev/esphome/components/esp32/__init__.py#L136) and so that board name does not work with ESPHome.

jesserockz:
More specifically, until the ESPHome core code is upgraded, there is no point in adding these to the dashboard UI as it will just cause issues for people who choose it.

lakshanthad:
I understand. When will you upgrade the ESPHome core code with the latest version?

PR #349, opened 2022-10-21
Remove position fixed from header & rename header element from div to header

Misiu:
I've reordered items in the layout (header above main), renamed div to the header, and removed fixed position because it is not needed with flex.

### DIFF
--- a/raw_package/static/css/esphome-2.css
+++ b/raw_package/static/css/esphome-2.css
@@ -41,6 +41,7 @@ body {
 
 /* Layout */
 main {
+  margin-top: 56px;
   flex: 1 0 auto;
   display: grid;
 }
@@ -52,11 +53,16 @@ esphome-main {
 }
 
 .esphome-header {
+  position: fixed;
+  top: 0;
+  right: 0;
+  left: 0;
   display: flex;
   align-items: center;
   height: 56px;
   background-color: var(--esphome-background-header);
   padding: 0 24px;
+  z-index: 1;
 }
 
 .flex {
@@ -122,7 +128,6 @@ ul.browser-default li {
   display: flex;
   align-items: center;
   min-height: 50px;
-  padding-top: 0;
   color: grey;
   background-color: var(--primary-footer-bg-color);
 }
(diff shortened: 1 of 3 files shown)


Misiu:
I didn't notice that. I'll revert position fixed

Misiu:
@balloob done. I had to add z-index to have the header above the cards.

balloob:
So the end result of this PR is that nothing has changed? 😄

raw_package/static/css/esphome-2.css
```diff
@@ -41,7 +41,7 @@ body {
 
 /* Layout */
 main {
-  margin-top: 64px;
+  margin-top: 56px;
```
balloob, line 44: does this mean all the content is 8px closer to the header. Do we compensate that with a padding in the content ?

Misiu:
> So the end result of this PR is that nothing has changed? 😄

I've renamed div to header and reordered the items. So a tiny change :)
I want to clean things up before I start adding CSS variables.

raw_package/static/css/esphome-2.css
Misiu: The padding isn't needed. The main element has space on the top (above the cards). I'll add screenshot showing before and after.

balloob: I think you forgot to add the screenshot?

Misiu: I'm away from my pc, will add today when I'm home.

Misiu:
This is how the change looks like. 64px (before) and 56px (after)
https://user-images.githubusercontent.com/1741838/197403200-00e0d263-e772-4254-a1c3-212c2e6cd8f1.mp4

balloob:
I think that 64px looks nicer as it's equal left/top/right.

Misiu:
> I think that 64px looks nicer as it's equal left/top/right.

the grid inside esphome-devices-list is 90% width of the screen, so I don't see the consistency.
Take a look at screenshot on wider screen:
image

PR #353, opened 2022-10-28
Gitpod

alexiri:
Add support for quick development environment creation with Gitpod.


balloob:
Not interested

alexiri:
May I ask why not? There is already partial support in [esphome](https://github.com/esphome/esphome/blob/dev/.gitpod.yml), though it's actually broken (would be fixed by https://github.com/esphome/esphome/pull/3962).

PR #358, opened 2022-11-09
Sort boards by name

Misiu:
Sort options in optgroup by value (case insensitive).
I've used this small code: https://codesandbox.io/s/sort-items-6d8mkb
Closes https://github.com/esphome/dashboard/issues/356


src/wizard/boards.ts
```diff
@@ -2,23 +2,23 @@ import { html } from "lit";
 
 export const boardSelectOptions = html`
   <optgroup label="ESP32">
+    <option value="featheresp32">Adafruit ESP32 Feather</option>
     <option value="esp-wrover-kit">Generic ESP32 (WROVER Module)</option>
```
balloob, line 6: Generic needs to stay on top. 

Let's just only sort the "Other" groups, the other ones are small and we sort them manually by popularity.

Misiu: Done. I've restored the sort in ESP32 and ESP8266 groups.

balloob (approved):

PR #359, opened 2022-11-09
Search

Misiu:
_This is just a draft because I want some initial feedback and I have some things to fix._

This PR adds a search box on top of the device list. It allows searching configured devices by name and comment properties.

esphome_search

Things added:
-toggle search button to header menu
-search input with autofocus
-clear search with escape key or on hide
-replaced buttons with icons in header menu
-added icons to header menu on mobile
-decrease margin in list menu item between text and icon

Closes https://github.com/esphome/feature-requests/issues/1888

Things to consider:
-where to add the search box?
-should it fix fixed on top? Currently, when we have many devices it will hide on scroll
-should the search use both properties (name, comment) or maybe only name? or maybe also the configuration
-how to fix the animation (highlight) problem?
-should we also add a combobox to allow filtering by status?
-how can we share components between HA and ESPHome? Because now I had to copy/paste almost 1:1 a couple of components.


randybb:
And what about that "feature" (which I hate and I never thought I will ever propose to somebody) that can capture (steal) ctrl+f / control+f and instead of buildin search in the browser cursor will focus to this one?

Misiu:
@randybb I'm trying to replicate the search that is available in HA.
I'm unsure if I should highlight the items the user is searching for or remove items that don't match the criteria (that's what I have so far).
one thing to notice: ctrl+f won't work on mobile devices

randybb:
Then hide it into that three dot menu?

Misiu:
> Then hide it into that three dot menu?

I don't want to hide them permanently. I just want to hide the items that don't match the search term. If I clear the search all items should appear normally. Take a look at gif I added to PR description.

randybb:
I was talking about "where to add the search box?".

I think this Search feature should be Filter - so it will remove all items which doesn't match criteria. If you want to jump to an iterm, classic search, then the user can just use buildin search in his browser.

Misiu:
> I was talking about "where to add the search box?".
> 
> I think this Search feature should be Filter - so it will remove all items which doesn't match criteria. If you want to jump to an iterm, classic search, then the user can just use buildin search in his browser.

So a third option (next to `UPDATE ALL` and `SECRETS`) on the desktop, and on the mobile in the 3 dots menu, it should toggle the search bar visibility. Nice idea!

I'll wait for @balloob opinion and will move forward with this.

Misiu:
I've added a button to toggle the search input, it also is autofocused when toggled and can be cleared with the escape key.

balloob:
I am not too convinced we need a search box and I’m afraid we’re cluttering the UI. Since we render all cards, a user can just use browser search to achieve the same? And if cards are ordered alphabetically, it’s also quick to scroll to the right place.

randybb:
search is not needed, as it is buildin in any browser. but filtering might be handy.

Misiu:
It was requested in a couple of places (FR, forum, and WTH), and yes, you can use browser build-in search, but for me, this is a filter in some way:
If you have multiple devices and only want to see lights related devices you search for `light` and the list is filtered. You can call devices as you like, so they don't have to be one after another on the dashboard.
Ideally, we can assign tags to devices and then filter by status and tags (labels to be consistent with the [HA Core PR](https://github.com/home-assistant/core/pull/69996)), but I think we don't have tags that we could assign to devices and use to filter the list.
Any thoughts on that?

Misiu:
I'm not trying to force changes, just trying to understand the reasoning.
I'll try to add a filter if that's more useful.

randybb:
Then this should be removed too as users could use search in their browser.

obrázok

craigueh:
Could this search function also search by IP address? This would be a mega help. For example, I have a lot of Sonoff/shelly devices and when adding new ones or replacing them I often struggle to see which device on the network is the device in the esphome dashboard. I can easily get the ip address of the device from my Unifi network controller but then have to manually view the logs of every suspected esphome device to find which one has the IP I'm looking for. 
If a search by IP was possible this would be so easy. Or at the very least, also show the device IP under the device name for each node in the Dashboard and then just use a browser text search for the IP I'm hunting for.

randybb:
1. No, this is not going to be added.
2. Don't want to see their IP addresses in my dashboard - I use DHCP and hostnames everywhere (dns, mdns) so it is useless information.

craigueh:
If it was useless information I wouldn't have asked for it. There are situations where it it is very useful to be able to determine which device is which by IP address. It's well known that dns/mdns is not always accurate/reliable. DHCP reservations are far from bullet proof too. Especially where Ubiquite networks are concerned. 

Take it from someone who's worked in the IT industry for 25+ years. Being able to pinpoint a device by IP address when DNS starts acting up is part of our daily routine whether we like it or not. It's hardly a regression to be able to put either an IP or hostname into a search box and have it return the correct relevant device.

Misiu:
I'll try to create a second version of this PR, including filter functionality (by status) and maybe search.
I don't mind adding an option to search by all properties that esphome is returning to the dashboard (especially by those that aren't displayed and can't be searched via browser build-in search)

This is a sample data that I get fro my test server:
```
  {
    "name": "Garage sensor",
    "configuration": "garage_sensor.yaml",
    "loaded_integrations": [
      "api",
      "async_tcp",
      "captive_portal",
      "esp32",
      "esphome",
      "logger",
      "md5",
      "mdns",
      "network",
      "ota",
      "preferences",
      "socket",
      "web_server_base",
      "wifi"
    ],
    "deployed_version": "2022.8.0",
    "current_version": "2022.8.1",
    "path": "/config/test.yaml",
    "comment": "garage motion + temperature sensor",
    "address": "test.local",
    "web_port": null,
    "target_platform": "ESP32"
  },
```

As you can see I get the address field (mdns), but no IP. Please check this on your side, if you also don't have the IP address, then create PR (https://github.com/esphome/feature-requests/issues). If the devs add that field and my next pr will get approved then there is a chance to have a filter/search experience.

randybb:
We want to have this interface clear. ESPHome Dashboard is basically only for flashing, it is not a monitoring tool. If you need it so badly, you can send info from IP address sensor to HA or Unifi integration can provide exactly this information. Or just create a script which will fit your needs.

glmnet:
I would accept a node list explorer in the vscode extension

RebbePod:
@Misiu Did you make a PR for the filtering option?

Adding to the search the ability to type in "online", "offline" or "update" can be really useful

PR #360, opened 2022-11-14
add missing scripts to package.json

Misiu:
This change adds missing scripts to package.json.
This allows running develop from VS Code menu:
image

I've also added instructions for Windows users on how to set bash as a default shell per project to be able to call scripts.
Ref: https://stackoverflow.com/questions/74391024/vs-code-how-to-specify-npm-run-shell-per-project

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,14 @@ Check out this repository, run `npm install --legacy-peer-deps` and then run
 
 ```
 script/develop
-````
+```
+
+if you are developing on Windows machine using VS Code, you must set bash as a default shell.
+You can do that by creating `.npmrc` file in the root of the project with the following content:
+
+```
+script-shell=C:\Program Files\git\bin\bash.exe
+```
 
 It will start the dev server and will automatically re-bundle updated JavaScript (except for the `static` folder).
 
--- a/package.json
+++ b/package.json
@@ -7,7 +7,10 @@
   "author": "ESPHome maintainers",
   "license": "Apache-2.0",
   "scripts": {
-    "prepublishOnly": "script/build"
+    "develop": "script/develop",
+    "build": "script/build",
+    "develop_web": "web.esphome.io/script/develop_web",
+    "build_web": "web.esphome.io/script/build_web"
   },
   "devDependencies": {
     "@rollup/plugin-commonjs": "^23.0.2",


package.json
```diff
@@ -7,7 +7,10 @@
   "author": "ESPHome maintainers",
   "license": "Apache-2.0",
   "scripts": {
-    "prepublishOnly": "script/build"
```
balloob, line 10: Do not remove prepublishOnly. That is used to ensure we have a build before publishing

package.json
```diff
@@ -7,7 +7,10 @@
...
+    "develop": "script/develop",
+    "build": "script/build",
```
balloob, line 11: Why would we add these here? They are never used except for on CI?

Misiu: by CI maybe yes, but I use them from VS Code to start dev servers:
image

balloob: You don’t ever call the build script though. 

Also, isn’t that provided by a plugin? I don’t have those.

package.json
balloob: Google it?

package.json
Misiu: It is built-in functionality: https://code.visualstudio.com/updates/v1_23#_npm-script-running.
I've added all scripts because now you can use npm to run them.
In the past, I did some things in React, and by default when creating new projects all actions are added to the scripts section in package.json.

README.md
```diff
@@ -10,7 +10,14 @@ Check out this repository, run `npm install --legacy-peer-deps` and then run
 
 ```
 script/develop
-````
+```
+
+If you are developing on Windows machine using VS Code, you must set bash as a default shell to be able to call the above script.
```
balloob, line 15: We shouldn’t include this. This is very specific to how you have bash installed (as part of how you have git installed). 

On Windows people should use the Windows Linux subsystem for development.

Misiu: There aren't any instructions on how to develop on Windows, you can't assume every developer is using Linux or macOS.
I've tried WSL before and I got errors when trying to run dev env. Setting bash as a default shell on Windows solved the problem for me and I could start contributing.
I agree that a dev container (as in HA Frontend repo) would be probably better, but I have no experience setting it up.

I can remove changes in Readme, but I think that for now, we should describe how to set up everything on Windows so that new contributors can start developing and debugging. Ideally, we should have a dev container, but maybe someone from HA Frontend team could help us with this?

balloob: We cannot add instructions for every permutation of development environments that people have.

The way you install git, where it is installed, if it includes bash (a different tool) is all based on assumptions. It just means we’re having to add a never-ending list of ways people might be developing. 

Instead, Microsoft has developed WSL so it is possible to have a single way of development across platforms.

Misiu: I've tried WSL (Ubuntu on Windows) and it didn't work for me, I'm getting errors all the time:
image

I can remove that part of the change (or you can do it), but I think you will agree that the readme should be improved.
Maybe a dev container? That would solve the getting started with development issue.

Misiu: I think I fixed the errors. One was an old nodejs version, and the second was the wrong end-line characters in scripts. It was CRLF, but it needs to be LF.

maybe instead of these changes to the readme, I can add a link to https://code.visualstudio.com/docs/remote/wsl?

glmnet: What about adding `#!/bin/bash` to scripts. That will make it clear you need bash

Misiu: good idea. I've added it to every script and changed end line character (at least now this works in WSL for me)
@glmnet any experience with dev containers?

Misiu: @balloob I don't want to waste anyone's time. I contributing my free time to this project, but I'm still learning web components and front-end stuff (I used GitHub actions and npm scripts to deploy applications, but I never used `prepublishOnly` option before, once again sorry for renaming it). I'm a backend developer, using Windows from the first day I got my PC. I think I don't have to know everything to help, don't you agree?

Leaving that behind I changed the end line because the scripts that git downloaded didn't work for me in WSL, I had to change them to stop getting errors. HA Frontend is using https://github.com/home-assistant/frontend/blob/dev/.gitattributes so that probably solves that problem (every file in script folder in HA frontend has LF as end line)

I'll roll back changes to readme, should I also rollback change with end of line character?

balloob: Sorry, that came off stronger than I intended. 

Changes are good here. Let's merge it.

Misiu: No worries :) I'm really glad that you help me on this bumpy road

PR #374, opened 2022-12-06
Add ESP32-S3 Support

ntwb:
This PR seeks to add _core_ dashboard support for ESP32-S3 devices

The PR is based on #201 that added ESP32C3 and ESP32S2 support

To test run:

1. `ESPHOME_DASHBOARD_DEV=../dashboard esphome dashboard ./`
2. Click "+ New Device"
3. Observe that ESP32-S3 is now a selectable option from the primary menu

image

PR #375, opened 2022-12-06
Add ESP32-C3/S2/S3 Boards

ntwb:
Supplemental PR to #374 and this PR seeks to add ESP32-C3, ESP32-S2, and ESP32-S3 boards to the dashboard dropdown menu

Each type of board has been added to it's option-group rather than crowding the "Other ESP32s" option group:

ESP32-C3
- image

ESP32-S2
- image
- image
ESP32-S3
- image

The boards list was taken from the ESPhome board variants list and extracting the 3 variants:
- `VARIANT_ESP32c3`
- `VARIANT_ESP32S2`
- `VARIANT_ESP32S3`

https://github.com/esphome/esphome/blob/dbbbba3cf8ff5ad73f200c085f0a7f84c6be9aa1/esphome/components/esp32/boards.py#L1081

> **Note**
> One thing I haven't done in this PR is written the "friendly name" for each board variant and have at this stage used the same option value for the board name


src/wizard/boards.ts
```diff
@@ -157,4 +157,62 @@ export const boardSelectOptions = html`
     <option value="xinabox_cw02">XinaBox CW02</option>
     <option value="nscreen-32">YeaCreate NSCREEN-12</option>
   </optgroup>
+  <optgroup label="ESP32-C3">
+    <option value="adafruit_qtpy_esp32c3">adafruit_qtpy_esp32c3</option>
+    <option value="airm2m_core_esp32c3">airm2m_core_esp32c3</option>
+    <option value="bee_motion_mini">bee_motion_mini</option>
+    <option value="deneyapkartg">deneyapkartg</option>
+    <option value="dfrobot_beetle_esp32c3">dfrobot_beetle_esp32c3</option>
```
balloob, line 165: We should write-out all the names. If we don't do that, we might as well replace it all with an open input text box. Should be possible with some scraping of the platformio docs

src/wizard/boards.ts
```diff
@@ -157,4 +157,62 @@ export const boardSelectOptions = html`
...
+    <option value="esp32s3box">Espressif ESP32-S3-Box</option>
+    <option value="esp32s3camlcd">ESP32S3 CAM LCD</option>
+    <option value="lolin_s3">WEMOS LOLIN S3</option>
+    <option value="tamc_termod_s3">TAMC Termod S3</option>
+    <option value="um_feathers3">Unexpected Maker FeatherS3</option>
+    <option value="um_pros3">Unexpected Maker PROS3</option>
+    <option value="um_tinys3">Unexpected Maker TinyS3</option>
+    <option value="unphone8">unPhone 8</option>
+    <option value="unphone9">unPhone 9</option>
+    <option value="wifiduino32s3">Blinker WiFiduino32S3</option>
```
balloob, line 216: We should order the options by the names, not their values.

balloob:
Looks like to keep things in sync it's better to just have the backend be a source of truth. Closing this PR in favor of #376

PR #384, opened 2022-12-12
Hide files with ".hide." in the filename from the UI

myztillx:
Currently there is no way to hide files/cards from the frontend UI. When developing or working on configuration files, you may not want them displayed in the UI. This is an example of a way this could be done via the filename. Another way could be adding information directly in the file themselves and gotten from the devices endpoint.

Running "Update All" will still try and update the hidden files. I am not sure where the change would need to be made skip files ending with ".hide." in the name.

New to Lit (and dev in general), so please let me know if this sort of PR is not acceptable.

### DIFF
--- a/src/devices/devices-list.ts
+++ b/src/devices/devices-list.ts
@@ -67,18 +67,23 @@ class ESPHomeDevicesList extends LitElement {
         ${repeat(
           this._devices.configured,
           (device) => device.name,
-          (device) => html`<esphome-configured-device-card
-            ${animate({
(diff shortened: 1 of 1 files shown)


randybb:
Is it not easier just to rename it from _foo.yaml_ to _.foo.yaml_?

myztillx:
It is, I was unaware that was a functionality... It does make the organization of files by name a bit more difficult, but seems to achieve everything I was looking for! Sorry for the unnecessary pull request.

PR #401, opened 2023-01-02
Fix "Show API Key" dialog shows "Loading..." forever if no encryption key in config

### LINKED ISSUE #400: "Show API Key" dialog shows "Loading..." forever if no encryption key in config
Selecting "Show API Key" from this menu:
Screen Shot 2023-01-01 at 7 34 00 PM

Displays this dialog:
Screen Shot 2023-01-01 at 7 33 53 PM

I expected it to either fail with an error, or display a message that I have not yet set up an API key. Instead, it just looks like it hangs, because it never updates.

- Under api, I only have `api: { password: "..." }` in my config, and that successfully comes back from the `json-config` call.
- A javascript breakpoint shows this method executes, and (possibly?) results in `undefined` (based on the type signature) https://github.com/esphome/dashboard/blob/8d0b3feed786792fab9e46b17e5a7964a81633ec/src/api/configuration.ts#L82-L87
- I don't know whether `this._apiKey` is updated or not https://github.com/esphome/dashboard/blob/8d0b3feed786792fab9e46b17e5a7964a81633ec/src/show-api-key/show-api-key-dialog.ts#L64-L66
- if it is, and the value is `undefined`, this code looks like it would continue to render as 'Loading...' https://github.com/esphome/dashboard/blob/8d0b3feed786792fab9e46b17e5a7964a81633ec/src/show-api-key/show-api-key-dialog.ts#L26-L27

I'm not set up for dashboard development (maybe I'll poke at it later), but I wanted to file the issue. I didn't see an existing Issue for this bug

e28eta:
I think this is a minimally correct set of changes to fix #400. When choosing the "Show API Key" menu item, this will show "Loading..." until a response comes back, and then it either shows the key or an error message, based on whether we get a string or `null`.

The wizard-dialog doesn't differentiate between `undefined` and `null`, instead using truthiness of the `_apiKey` to show / hide a section with the key. In my testing, I wasn't able to find a path that'd trigger the "no API Key" flow in the wizard, but I don't think I've broken anything.

I don't have much experience with this web stack, so I hope I've done an okay job w/the changes.

### DIFF
--- a/src/api/configuration.ts
+++ b/src/api/configuration.ts
@@ -81,7 +81,7 @@ export const getJsonConfig = async (
 
 export const getConfigurationApiKey = async (
   configuration: string
-): Promise<string | undefined> => {
+): Promise<string | null> => {
   const config = await getJsonConfig(configuration);
-  return config?.api?.encryption?.key;
+  return config?.api?.encryption?.key || null;
 };
--- a/src/show-api-key/show-api-key-dialog.ts
+++ b/src/show-api-key/show-api-key-dialog.ts
@@ -26,8 +26,8 @@ class ESPHomeShowApiKeyDialogDialog extends LitElement {
         ${this._apiKey === undefined
           ? "Loading…"
           : this._apiKey === null
-          ? html`Unable to automatically extract API key. Open the configuration
-              and look for <code>api:</code>.`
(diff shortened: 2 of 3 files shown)


balloob (approved):

balloob:
Thanks 🐬

PR #429, opened 2023-02-05
First pass at a dark theme

goldbattle:
I was unable to get the full esphome stack running on my windows machine, but was able to host a small local web server with the static files in the [web.esphome.io](https://github.com/esphome/dashboard/tree/main/web.esphome.io) folder. This uses the user's `prefers-color-scheme: dark` setting to change some css values so it is a bit better on the eyes. This should close https://github.com/esphome/dashboard/issues/347.

Here are some screenshots:

screencapture-127-0-0-1-8080-2023-02-05-03_11_45

screencapture-127-0-0-1-8080-2023-02-05-03_12_16

I was thus unable to test any screens past these ones, so the editor will still be the light theme.
I believe this would require changing the following code.
Maybe this should be a config setting, or maybe [check with js](https://stackoverflow.com/a/57795495/7718197) and toggle to `vs-dark` with a darker `editor.foreground`?
https://github.com/esphome/dashboard/blob/e60c7d936de424b9360ab892317723c4277404af/src/editor/monaco-provider.ts#L57-L65

### DIFF
--- a/web.esphome.io/src/dashboard/unsupported-card.ts
+++ b/web.esphome.io/src/dashboard/unsupported-card.ts
@@ -23,6 +23,9 @@ class EWUnsupportedCard extends LitElement {
       esphome-card {
         --status-color: var(--alert-error-color);
       }
+      .card-actions mwc-button {
+        --mdc-theme-primary: var(--card-text-color);
+      }
     `,
   ];
 }
--- a/raw_package/static/css/esphome-2.css
+++ b/raw_package/static/css/esphome-2.css
@@ -31,15 +31,15 @@
     --primary-bg-color: #212121;
     --primary-footer-bg-color: #080808;
 
-    --mdc-theme-surface: #C0C0C0;
+    --mdc-theme-surface: #c0c0c0;
 
-    --esphome-background-header: #C0C0C0;
+    --esphome-background-header: #c0c0c0;
 
     --card-background-color: #888888;
-    --card-status-color: #FFFFFF;
+    --card-status-color: #ffffff;
   }
   html {
-    color: #FFFFFF;
+    color: #ffffff;
   }
 }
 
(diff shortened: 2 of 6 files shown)


Misiu:
@goldbattle I have some progress on the dark theme you might take a look: https://github.com/Misiu/esphome_dashboard/commit/6c8b82afd146e42280add18ab008cd17643ab4b6
Sadly not enough time to move this forward in the near future.

goldbattle:
Thanks for the pointer. I tried to change as little as possible here in this PR.

I am using esphome in home assistant, and I couldn't find that css file.
Is the `esphome-2.css` is shared between this and the website?

balloob:
This theme does not look good. I would suggest that you take the dark themes of Home Assistant and use those color values.

goldbattle:
Thanks, here are the images of what it looks like using the CSS `#111111` background, `#1c1c1c` card, and `#e1e1e1` text from the [frontend](https://github.com/home-assistant/frontend/blob/d15d339782b89dab0d6d9c029abafe5c140c0d63/src/resources/styles.ts#L3-L7) repo. Let me know any additional comments.

Page 1:
image
image


Page 2:
image
image

web.esphome.io/src/dashboard/unsupported-card.ts
```diff
@@ -23,6 +23,9 @@ class EWUnsupportedCard extends LitElement {
       esphome-card {
         --status-color: var(--alert-error-color);
       }
+      .card-actions mwc-button {
+        --mdc-theme-primary: var(--card-text-color);
```
balloob, line 27: It looks wrong to set a theme color to a text color variable.

goldbattle: This is what is currently done to change the color:
https://github.com/esphome/dashboard/blob/270f7e462bcf9e016f1169bca40548e091586ae8/web.esphome.io/src/dashboard/esp-connect-card.ts#L81-L83

In the docs it says "Color of a flat or outlined button. Background color of a raised button."
https://www.npmjs.com/package/@material/mwc-button#user-content-global-custom-properties

PR #475, opened 2023-04-20
Fix firmware URL

balloob:
Fix the URL where we download firmware from.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -4212,8 +4212,7 @@
     },
     "css-declaration-sorter": {
       "version": "6.3.1",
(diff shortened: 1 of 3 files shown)


bhyatyab:
Also getting similar error or pi2. Please fix url

Can't install ghcr.io/esphome/esphome-hassio:2023.4.2: 404 Client Error for http+docker://localhost/v1.42/images/ghcr.io/esphome/esphome-hassio:2023.4.2/json: Not Found ("No such image: ghcr.io/esphome/esphome-hassio:2023.4.2")

PR #496, opened 2023-06-19
Add ACP schema

grahambrown11:


balloob:
Should we wait for @glmnet to push a full schema update? https://github.com/esphome/esphome-vscode/issues/79#issuecomment-1599811154

glmnet:
I'm ok with this, it will be just be overwritten later

The docs still needs some updates for the schemas to be loaded properly

glmnet:
superseded by #499

PR #497, opened 2023-06-19
Dashboard & Web Installer Dark Theme

grahambrown11:
Dark Mode / Theme for both Dashbord & Web Installer as they both use the same base CSS


grahambrown11:
I know there is an older PR trying to do the same, but as the web installer and dashboard share css and some components it had to be done for both, here are some screenshots:
image
image
image

raw_package/static/css/esphome-2.css
```diff
@@ -15,11 +14,52 @@
   --alert-warning-color-bg: #fff5e6;
   --alert-error-color: #d93025;
   --alert-error-color-bg: #faefeb;
-
-  --mdc-theme-primary: #03a9f4;
+  --mdc-theme-primary: #000000;
```
balloob, line 17: Any reason we're changing this to black ?

raw_package/static/css/esphome-2.css
```diff
@@ -15,11 +14,52 @@
...
+  --primary-text-color: #000000;
   --mdc-dialog-z-index: 998;
-
+  --mdc-theme-primary-no-attention: #444444;
+  --mdc-theme-on-primary-no-attention: white;
   --esphome-background-header: #e0e0e0;
+  --mdc-theme-on-primary: white;
+  --mdc-dialog-heading-ink-color: #000000;
+  --mdc-dialog-content-ink-color: #000000;
```
balloob, line 25:
```suggestion
  --mdc-dialog-heading-ink-color: #212121;
  --mdc-dialog-content-ink-color: #212121;
```

src/install-choose/install-choose-dialog.ts
```diff
@@ -480,6 +480,9 @@ class ESPHomeInstallChooseDialog extends LitElement {
       svg {
         fill: currentColor;
       }
+      .svg-filter {
```
balloob, line 483: Move this to `esphomeDialogStyles`. Let's also rename it to `svg-fill-primary`

web.esphome.io/src/components/ew-header-menu.ts
```diff
@@ -54,7 +54,6 @@ export class EWHeaderMenu extends LitElement {
 
   static styles = css`
     mwc-button {
-      --mdc-theme-primary: black;
```
balloob, line 57: You're changing this button to be a different color. Why?

balloob (changes requested):
Couple of comments.

raw_package/static/css/esphome-2.css
grahambrown11: As I removed the override in each button that was set to black, otherwise all button text is blue

web.esphome.io/src/components/ew-header-menu.ts
grahambrown11: In order for both light and dark to work, I've removed the override of the theme in each component, if I don't remove this the button text stays black in dark mode...
Without removing:
image
Removed:
image

nagyrobi:
@grahambrown11 also see this suggestion: https://github.com/esphome/feature-requests/issues/1003 (maybe for a later PR).

grahambrown11:
> @grahambrown11 also see this suggestion: [esphome/feature-requests#1003](https://github.com/esphome/feature-requests/issues/1003) (maybe for a later PR).

Currently as the dashboard is not aware of HA that'll be a complex change to get the users HA theme...
I'm hoping this'll work for most ppl who prefer dark mode they'll have it set in their OS...
Maybe allowing the user to chose a light / dark theme like some websites do and default to OS prefered mode would offer more flexibility... I'll see what I can do

nagyrobi:
As I said, don't stress yourself on this, I just linked it to see some of the (eventually upcoming) user demands...

src/install-choose/install-choose-dialog.ts
grahambrown11: I opted to create a new `esphomeSvgStyles` as there was use in the web installer that was not in a dialog

grahambrown11:
Added a user toggle for the theme incase the user wants to specify over the browsers `refers-color-scheme`
image
image

raw_package/static/css/esphome-2.css
```diff
@@ -15,11 +14,50 @@
   --alert-warning-color-bg: #fff5e6;
   --alert-error-color: #d93025;
   --alert-error-color-bg: #faefeb;
-
-  --mdc-theme-primary: #03a9f4;
+  --mdc-theme-primary: #212121;
```
balloob, line 17: I think it's a mistake. Instead, we should just override the button to have the theme set to the color that we want (`--primary-text-color`). If we ever want to use the theme primary on the material components, we wouldn't want to be stuck with black and it being a huge breaking change.

grahambrown11: Have reverted the root `--mdc-theme-primary`

src/styles.ts
```diff
@@ -107,4 +98,14 @@ export const esphomeDialogStyles = css`
...
   }
+
+  mwc-button {
+    --mdc-theme-primary: var(--primary-text-color);
+  }
+`;
+
+export const esphomeSvgStyles = css`
+  .svg-fill-primary {
+    fill: var(--primary-text-color);
```
balloob, line 109: I wonder if `fill: currentColor` isn't the better way to go here, since that makes the icons colored like the text.

balloob: this can be fixed in a follow-up PR. Let's get this in the beta.

balloob:
Found a bug, this button is purple now:

CleanShot 2023-06-20 at 21 57 30

balloob:
Ah nevermind, Jesse found out the CSS was cached for ESPHome Web. All good 👍

balloob:
The dropdown for picking wifi network is not styled. However, I think it's because it's actually the Improv Wi-Fi SDK that's embedded in the ESPHome Web that shows this dialog, not a big deal.

CleanShot 2023-06-20 at 22 10 00

nagyrobi:
> Feedback on discord is to not have this button

Where? Didn't see that...
I would have voted to have it.

grahambrown11:
Thank you @balloob for your review and helping getting this over the line 🎉 
I and the DrZzs community will like not having our retinas burnt when the doc goes to ESPHome now on his live streams 😺

golles:
Hi @grahambrown11 
I can't seem to create an issue in this repo, but I think you forgot to check the login screen, all the text and icons are dark on a dark background, see next picture:

Scherm­afbeelding 2023-06-24 om 14 17 23

grahambrown11:
> I think you forgot to check the login screen

Oh sorry, I will look into that now... I only use this in HomeAssistant, so have not seen that login mode...

grahambrown11:
@golles hoopefully my new PR will be accepted - this should sort it https://github.com/esphome/dashboard/pull/500/files#r1240875893

balloob:
Can we move the search into the top toolbar, just icon only, and have the search overlay the toolbar when active. That way it doesn't take up a lot of space when not used

balloob:
Search icon could still be shown next to the overflow menu icon on mobile.

cenobitedk:
> ~Added a user toggle for the theme incase the user wants to specify over the browsers `prefers-color-scheme`~ image image
> 
> Feedback on discord is to not have this button

@grahambrown11 What was the feedback, and for what reason?
As the web adopts standards for adapting to users systems preference, while still maintaining a mean to choose, I find it really strange (and not in a good way) to specifically choose not to follow standards.

grahambrown11:
> As the web adopts standards for adapting to users systems preference, while still maintaining a mean to choose, I find it really strange (and not in a good way) to specifically choose not to follow standards.

@cenobitedk The decision was not mine it was feedback from the maintainers of HomeAssistant & ESPHome. As for the standards, I'm not sure what you're talking of - there are no standards when it comes dark theme switches or the web in general, some websites chose to have a visible switcher some put it in user preferences but the majority don't even support it at all...

cenobitedk:
@grahambrown11 
>  The decision was not mine it was feedback from the maintainers of HomeAssistant & ESPHome. 

Yeah I know, I was curious about their reasoning.

> As for the standards, I'm not sure what you're talking of - there are no standards when it comes dark theme switches or the web in general, some websites chose to have a visible switcher some put it in user preferences but the majority don't even support it at all...

You're right that implementation various a lot, but there certainly is support for it on many many sites, and that is setting a de facto standard for other sites to follow.
There's no standard for login flow either, yet a billion sites follow the samme approach based on best practices in regards to security and UX.

Since we have wide spread browser support (https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) I doesn't make sense to go against the way other sites have implemented client support.

nagyrobi:
> Yeah I know, I was curious about their reasoning.

There was no reasoning (no democratic opinioning either) a question was raised, an a HA dev responded negative, in a short while that's it. No other users were listened. As soon as that single response came, this was done.

cenobitedk:
That's just sad that a single dev get's to decide what's good user experience. :-( and it's in opposition to how it's done in HA:
image

I think this deserves a revote. I think the good work @grahambrown11 has done deserves to be in the dashboard so that people can choose for themselves rather than what a single dev thinks.

PR #500, opened 2023-06-24
Search/Filter Feature

grahambrown11:
- Implement search
- Found some buttons in dialogs didn't get the text style from the dark mode refactoring
- Tweak the status logic so it has a offline status rather than no status at all
- Change the footer so web.esphome.io does not have a strange `vWEB Documentation` link

### DIFF
--- a/raw_package/base_with_header_footer.template.html
+++ b/raw_package/base_with_header_footer.template.html
@@ -41,13 +41,9 @@
     <footer class="page-footer">
       <div>
         ESPHome by Nabu Casa |
-        <a href="https://esphome.io/guides/supporters.html" target="_blank"
-          >Fund development</a
-        >
+        <a href="https://esphome.io/guides/supporters.html" target="_blank">Fund development</a>
         |
-        <a href="{{ docs_link }}" target="_blank" rel="noreferrer"
-          >v{{ version }} Documentation</a
-        >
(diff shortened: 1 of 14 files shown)


grahambrown11:
Some screnshots of search bar:
image
image

raw_package/login.template.html
grahambrown11, line 1: Here's a preview of these changes
image
image

balloob:
Get stack trace when searching. Searching is also very slow. 

CleanShot 2023-06-24 at 22 29 37

balloob:
There should be a message shown when the search didn't match any devices, now the page is just empty.

balloob:
Can we move the search into the top toolbar, just icon only, and have the search overlay the toolbar when active. That way it doesn't take up a lot of space when not used. 
Search icon could still be shown next to the overflow menu icon on mobile.

grahambrown11:
@balloob 
> Get stack trace when searching. 

Can I add source maps so we get a more useful info... dunno where that is...

> Searching is also very slow.

I'll see what I can do to improve it - I think Lit is not rendering immediately on the change of the search input... (I am used to Angular)

> There should be a message shown when the search didn't match any devices

Will add message

> Can we move the search into the top toolbar, just icon only

I can, but wouldn't that UX be potentially confusing if the input is hidden but there is a value that is filtering the results... (also being selfish that's more work than I bargained for :wink:)

PR #512, opened 2023-07-18
streamer mode fix

grahambrown11:
Toggle the streamer mode in the add-on config to protect your secrets:
image

### DIFF
--- a/raw_package/index.template.html
+++ b/raw_package/index.template.html
@@ -8,6 +8,7 @@
 <esphome-main
   version="{{ version }}"
   docsLink="{{ docs_link }}"
+  {% if streamer_mode %}streamerMode="true"{% end %}
   {% if login_enabled %}logoutUrl="{{ relative_url }}logout"{% end %}
 ></esphome-main>
 {% end %}
--- a/src/components/remote-process.ts
+++ b/src/components/remote-process.ts
@@ -28,7 +28,7 @@ class ESPHomeRemoteProcess extends HTMLElement {
         }
         ${coloredConsoleStyles}
       </style>
-      <div class="log"></div>
+      <div class="log${ColoredConsole.blurSecrets ? " blur-secrets" : ""}"></div>
     `;
 
     const coloredConsole = new ColoredConsole(shadowRoot.querySelector("div")!);
--- a/src/editor/esphome-editor.ts
+++ b/src/editor/esphome-editor.ts
@@ -1,6 +1,6 @@
 import * as monaco from "monaco-editor/esm/vs/editor/editor.api";
 import { LitElement, html } from "lit";
-import { customElement, property, query } from "lit/decorators.js";
+import { customElement, property, query, state } from "lit/decorators.js";
 import "@material/mwc-dialog";
 import "@material/mwc-button";
 import "@material/mwc-snackbar";
@@ -12,6 +12,7 @@ import type { Snackbar } from "@material/mwc-snackbar";
 import { fireEvent } from "../util/fire-event";
 import { debounce } from "../util/debounce";
 import "./monaco-provider";
+import { mdiIncognito } from "@mdi/js";
 
 // WebSocket URL Helper
 const loc = window.location;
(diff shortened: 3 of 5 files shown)


grahambrown11:
Resolves one part of https://github.com/esphome/issues/issues/4709

src/components/remote-process.ts
```diff
@@ -28,7 +28,9 @@ class ESPHomeRemoteProcess extends HTMLElement {
         }
         ${coloredConsoleStyles}
       </style>
-      <div class="log"></div>
+      <div class="log${
```
balloob, line 31: Use the `classMap` helper from Lit.

src/editor/esphome-editor.ts
```diff
@@ -34,19 +35,22 @@ class ESPHomeEditor extends LitElement {
...
-  protected render() {
-    const isSecrets =
-      this.fileName === "secrets.yaml" || this.fileName === "secrets.yml";
+  private _isSecrets() {
+    return this.fileName === "secrets.yaml" || this.fileName === "secrets.yml";
+  }
 
-    return html`
+  protected render() {
+    const commonStyle = html`
```
balloob, line 53: Move this outside the class definition.

src/editor/esphome-editor.ts
```diff
@@ -70,6 +74,60 @@ class ESPHomeEditor extends LitElement {
           white-space: nowrap;
           text-overflow: ellipsis;
         }
+      </style>
+    `;
+    const isSecrets = this._isSecrets();
+    if (isSecrets && this.streamerMode && !this._showSecrets) {
```
balloob, line 80: This feels weird to put it in this component. Instead, this dialog should be it’s own code that is called from the “Secrets” button. When user confirms, it will continue to open the editor.

src/util/console-color.ts
```diff
@@ -10,6 +10,8 @@ interface ConsoleState {
 }
 
 export class ColoredConsole {
+  public static blurSecrets = false;
```
balloob, line 13: This doesn’t seem to be used ?

raw_package/index.template.html
```diff
@@ -8,6 +8,7 @@
 <esphome-main
   version="{{ version }}"
   docsLink="{{ docs_link }}"
+  {% if streamer_mode %}streamerMode="true"{% end %}
```
balloob, line 11: Boolean properties are set to True based on their existence. 
```suggestion
  {% if streamer_mode %}streamerMode{% end %}
```

src/util/console-color.ts
```diff
@@ -24,7 +26,9 @@ export class ColoredConsole {
   constructor(public targetElement: HTMLElement) {}
 
   logs(): string {
-    return this.targetElement.innerText;
+    const logs = this.targetElement.cloneNode(true) as HTMLElement;
+    logs.querySelectorAll(".log-secret").forEach((e) => (e.innerHTML = ""));
```
balloob, line 30: Why is this necessary ?

src/util/console-color.ts
```diff
@@ -281,4 +287,13 @@ export const coloredConsoleStyles = `
   .log-bg-white {
     background-color: rgb(255, 255, 255);
   }
+  @media not screen {
```
balloob, line 290: Is this for when the page is printed?

src/editor/esphome-editor.ts
grahambrown11: That could be an approach, but this will work for when I've implemented routing the secrets page will have it;s own url `/secrets` the button will be a link to the page not a change in state...

src/components/remote-process.ts
grahambrown11: will do (new to lit, was not aware of this functionality)

src/editor/esphome-editor.ts
grahambrown11: 👍🏻

src/util/console-color.ts
grahambrown11: It is not used in this class, but is used as a global variable set in  esphome-main and used in remote-process I see that is confusing - I'll move it to remote-process rather.
It started out there as I wanted to set the style in color-consol, but the styles were created too early so it didn't update

src/util/console-color.ts
grahambrown11: When you used the download logs buttons then you get:
`BSSID: XX:XX:XX:XX:XX:XX[redacted]`
(I've redacted the actual BSSID the X's are an example)

This change is so that it downloads as:
`BSSID: [redacted]`

I've seen a few people share their logs this way we should protecting them 🙈

src/util/console-color.ts
grahambrown11: yes, just incase someone uses the print this will to show `[redacted]` instead of the secret...

src/components/remote-process.ts
grahambrown11: I had to revert the `classMap` change as this code is attaching to a shadow dom and lit directives are not supported in that... I got `class="[Object object]"` instead of the strings...

src/editor/esphome-editor.ts
balloob: We can always attach an event listener to the link to avoid it linking in streamer mode.

src/components/remote-process.ts
balloob: Lit directives do work in shadow dom. Their team published the first shadow dom framework :-) 

https://lit.dev/docs/templates/directives/#classmap

src/util/console-color.ts
grahambrown11: ok

src/components/remote-process.ts
grahambrown11: @balloob I'm not sure why but the `ESPHomeRemoteProcess` does not follow the usual lit patterns
It extends `HTMLElement` not `LitElement` and uses native `attachShadow` and `innerHTML` functions so the lit directive for the `classMap` does not work:

```typescript
    const shadowRoot = this.attachShadow({ mode: "open" });
    const classes = { log: true, "blur-secrets": ESPHomeBlurSecrets.enabled };
    shadowRoot.innerHTML = `
      <style>
        :host {
          display: flex;
        }
        ${coloredConsoleStyles}
      </style>
      <div class="${classMap(classes)}"></div>
    `;
```
produces:
```html
<div class="[object Object]">
```
which looks like this
image

Whereas the current implementation with the ternary operator produces the correct result:
image

src/editor/esphome-editor.ts
grahambrown11: I don't understand why you're not ok with this being in the editor.
There was already logic for the editor to handle secrets differently (no install & ace websocket not started)
so I don't see why it cannot be given 1 extra parameter (streamer mode) in order to show the warning 1st

src/components/remote-process.ts
balloob: Ah right. Yeah I don't remember why.

src/editor/esphome-editor.ts
balloob: A component should do 1 thing. This makes it very complicated with doing completely different things. Also, it's a prompt to a user so we should follow the convention of asking that as a dialog.

grahambrown11: OK created a dialog in `streamer-warning.ts`
image

grahambrown11: @balloob are you happy with this change?

src/components/remote-process.ts
```diff
@@ -3,6 +3,10 @@ import { StreamError, streamLogs } from "../api";
 import { ColoredConsole, coloredConsoleStyles } from "../util/console-color";
 import { fireEvent } from "../util/fire-event";
 
+export class ESPHomeBlurSecrets {
```
balloob, line 6: Why is this inside `remote-process` ? This can be in it's own file and doesn't have to be a class

src/editor/esphome-editor.ts
```diff
@@ -101,9 +112,24 @@ class ESPHomeEditor extends LitElement {
             ></mwc-button>`}
       </div>
       <main></main>
+      ${isSecrets && this.streamerMode && !this._showSecrets
+        ? html`<streamer-warning
```
balloob, line 116: It should be it's own dialog like how we open the wizard. The editor shouldn't know it's opening a secret. Just the header should, if streamer mode enabeld, prompt to confirm the user wants that before opening the editor. The editor should not be aware of streamer mode.

src/components/remote-process.ts
grahambrown11: I din't think it needed its own file as it is so small - and it needs to be a class as it is used as a static/global variable, I find this a better approach than using the `window` object to set globals...

src/editor/esphome-editor.ts
grahambrown11: The editor is already aware of secrets - thats not something new...
And as mentioned before, I think this is better here for when the secrets have its own URL (the other part I am working on), the header wont be able to show a popup if the user went directly to /secrets...

balloob: The only way a user can go to secrets is by pressing the button in the top toolbar. If they go any other way, streamer mode does not have to guard for that. The dialog needs to be it's own file.

PR #528, opened 2023-09-08
Small fixes

Misiu:
While browsing the code in VS I noticed some small places that showed errors when I removed `ignoreDeprecations` in tsconfig.

I've added the missing semicolon in src/devices/configured-device-card.ts and also adjusted the type property in src/components/process-dialog.ts and src/components/remote-process.ts
There was a missing cast in src/wizard/wizard-dialog.ts, so I've also added it.

### DIFF
--- a/src/install-choose/install-choose-dialog.ts
+++ b/src/install-choose/install-choose-dialog.ts
@@ -267,7 +267,7 @@ class ESPHomeInstallChooseDialog extends LitElement {
           <mwc-circular-progress
             active
             ?indeterminate=${progress === undefined}
-            .progress=${progress !== undefined ? progress / 100 : undefined}
+            .progress=${progress !== undefined ? progress / 100 : 1}
             density="8"
           ></mwc-circular-progress>
           ${progress !== undefined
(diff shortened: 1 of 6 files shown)


src/install-choose/install-choose-dialog.ts
```diff
@@ -267,7 +267,7 @@ class ESPHomeInstallChooseDialog extends LitElement {
           <mwc-circular-progress
             active
             ?indeterminate=${progress === undefined}
-            .progress=${progress !== undefined ? progress / 100 : undefined}
+            .progress=${progress !== undefined ? progress / 100 : 1}
```
balloob, line 270: This should be using the `ifDefined` helper from lit.

Misiu: I've tried adding ifDefined, but I get this error then:
image

not sure if this is a false-positive, but lit-plugin shows me many errors, including `Type 'true' is not assignable to 'string'` for empty scrimClickAction, ref: https://github.com/esphome/dashboard/pull/303

PR #540, opened 2023-10-23
remove use of npm legacy peer deps

glmnet:
Completes #525 as the 
`--legacy-peer-deps` option is still in use in github build actions


balloob (approved):
Thanks

PR #544, opened 2023-11-13
Prepare for ESP32-C6 support

jesserockz:
This adds the C6 to the supported platforms "list" and also adds a new key/value for removing items from the wizard. 

I started by adding to the `SupportedPlatforms` so that `chipFamilyToPlatform` could map it, but then typescript said I had to implement in the `supportedPlatforms` object too.

So in terms of user facing changes, it is not shown in the wizard yet, but will allow web-flashing of a C6 configuration.

### DIFF
--- a/src/const.ts
+++ b/src/const.ts
@@ -15,7 +15,7 @@ export type SupportedPlatforms =
 export type PlatformData = {
   label: string;
   showInPickerTitle: boolean;
-  showInDeviceType: boolean;
+  showInDeviceTypePicker: boolean;
   defaultBoard: string | null;
 };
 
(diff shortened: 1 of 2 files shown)


balloob:
Why not add support for it in the wizard ?

jesserockz:
Because until ESPHome updates the esp-idf and platformio versions to be at a minimum where C6 support was added, it just doesnt work :see_no_evil: 

It requires the user to manually change those versions in yaml until such a time.

src/const.ts
```diff
@@ -9,11 +9,13 @@ export type SupportedPlatforms =
   | "ESP32S2"
   | "ESP32S3"
   | "ESP32C3"
+  | "ESP32C6"
   | "RP2040";
 
 export type PlatformData = {
   label: string;
   showInPickerTitle: boolean;
+  showInDeviceType: boolean;
```
balloob, line 18: Let's change this `showInDeviceTypePicker` maybe?

balloob (approved):

ximex:
so esp32-c6 could be enabled after this change is released? https://github.com/esphome/esphome/pull/7706

randybb:
yup, and probably H2 (needs 5.1) and P4 (needs 5.2 => as hidden) can be added to the picker as well.

ximex:
created a PR now to enable ESP32-C6: https://github.com/esphome/dashboard/pull/711

PR #596, opened 2024-03-17
Add Link to ChangeLog

sebcaps:
Handle [#2623 ](https://github.com/esphome/feature-requests/issues/2623)to add a link to ChangeLog in footer.

### DIFF
--- a/src/esphome-main.ts
+++ b/src/esphome-main.ts
@@ -53,6 +53,8 @@ class ESPHomeMainView extends LitElement {
           <a href=${this.docsLink} target="_blank" rel="noreferrer"
             >${this.version} Documentation</a
           >
+          |
+          <a href="https://github.com/esphome/esphome/releases" target="_blank">Release Note</a>
         </div>
       </footer>
     `;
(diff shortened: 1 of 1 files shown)


src/esphome-main.ts
```diff
@@ -53,6 +53,8 @@ class ESPHomeMainView extends LitElement {
           <a href=${this.docsLink} target="_blank" rel="noreferrer"
             >${this.version} Documentation</a
           >
+          |
+          <a href="https://github.com/esphome/esphome/releases" target="_blank">Change Log</a>
```
jesserockz, line 57: This might be better
```suggestion
          <a href="https://esphome.io/changelog/" target="_blank">Change Log</a>
```

balloob:
I am not a fan of this change. The update entity in Home Assistant will list the changelog and the link at the bottom already links to the website, which will notify you directly of the latest changelog using a cookie.

wuppiwuppi:
some ppl do not use the dashboard directly in home assistant - rather standalone ...

balloob:
We don't want to implement this.

PR #598, opened 2024-03-18
Add CSRF token to requests when available in cookies

jesserockz:
See https://www.tornadoweb.org/en/branch6.4/guide/security.html#cross-site-request-forgery-protection

### DIFF
--- a/src/api/index.ts
+++ b/src/api/index.ts
@@ -1,3 +1,5 @@
+import { getCookie } from "../util/cookie";
+
 export class APIError extends Error {
   constructor(
     message: string,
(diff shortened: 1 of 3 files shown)


src/api/index.ts
```diff
@@ -16,6 +18,13 @@ const fetchApiBase = async (
     options = {};
   }
   options.credentials = "same-origin";
+  let csrfCookie = getCookie("_xsrf");
```
balloob, line 21:
```suggestion
  const csrfCookie = getCookie("_xsrf");
```

balloob (approved):

src/api/index.ts
jesserockz: Thanks

PR #600, opened 2024-03-26
feat: add support multiple device for name_add_mac_suffix on GUI

magicbear:
add support multiple device for name_add_mac_suffix on GUI
will be show a select for multiple devices

### DIFF
--- a/src/api/ota-ports.ts
+++ b/src/api/ota-ports.ts
@@ -1,0 +1,12 @@
+import { fetchApiJson } from ".";
+
+export interface ServerSerialPort {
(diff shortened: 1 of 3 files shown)


balloob:
We don't want to merge this. Configurations are tied to devices in the dahsboard. This PR is decoupling that and adds another step for the main flow of ESPHome device update. Advanced users can already achieve this from the ESPHome CLI. 

We're in the final phase of adding OTA update to devices without requiring the dashboard. This should be the way for firmware to update themselves without having a specific config.

PR #606, opened 2024-04-09
Get rid of horizontal scrollbar while searching

mkaatman:
Before:
image

After:
image

### DIFF
--- a/src/components/esphome-search.ts
+++ b/src/components/esphome-search.ts
@@ -29,8 +29,7 @@ export class ESPHomeSearch extends LitElement {
     css`
       esphome-text-input {
         --mdc-theme-primary: var(--primary-text-color);
-        width: calc(100% + calc(100vw - 100%));
-        max-width: 100%;
+        width: 100%;
       }
       .mdc-text-field__icon {
         color: var(--primary-text-color);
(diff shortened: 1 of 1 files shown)


mkaatman:
It's not pretty but this was the only way I could find to get rid of the horizontal scroll that appears when you search. Feel free to close if there's a better way.

For me the computed width without the change is 3220 and with the change is 3203.

src/components/esphome-search.ts
```diff
@@ -29,7 +29,8 @@ export class ESPHomeSearch extends LitElement {
     css`
       esphome-text-input {
         --mdc-theme-primary: var(--primary-text-color);
-        width: 100vw;
+        width: calc(100% + calc(100vw - 100%));
```
balloob, line 32: The problem is the scroll bar, but shouldn't just 100% width solve that?

mkaatman: Unfortunately no. The scrollbar persists. I think it's because of inherited margins or padding from a parent container. I'll ask the UX folks tomorrow at work if there's a cleaner way to solve it.

mkaatman: I stand corrected. I swear I tried that last night but it's working fine now. Adjusted PR accordingly.

balloob (approved):

PR #684, opened 2024-11-29
Fix parsing of blocks

swoboda1337:
Fix issues with parsing blocks:
- Right now every line needs to have the same indent. However C++ code inside a lambda will have varying indents. Instead of checking for a fixed indent check that the indent is greater than or equal to the indent on the first line. End the block only when the indent decreases. 
- The rest of the editor handles tabs except for the lambda/block part. Check for both tabs and spaces. 
- Allow whitespace after the block indicator. If there is a trailing space its not treated as a block and gets parsed as yaml.

Not only does the current parsing look bad it also confuses people into thinking indents in lambdas are not allowed. I have seen people post huge lambdas in discord without a single indent. 

Current HA addon, no whitespace after |-
image

Current HA addon, with whitespace after |-
image

Fixed
image

### DIFF
--- a/src/editor/monaco-provider.ts
+++ b/src/editor/monaco-provider.ts
@@ -265,16 +265,16 @@ monaco.languages.setMonarchTokensProvider("esphome", {
     ],
 
     // First line of a Block Style
-    multiString: [[/^( +).+$/, "string", "@multiStringContinued.$1"]],
+    multiString: [[/^([ \t]+).+$/, "string", "@multiStringContinued.$1"]],
 
     // Further lines of a Block Style
     //   Workaround for indentation detection
     multiStringContinued: [
       [
-        /^( *).+$/,
+        /^([ \t]*).+$/,
         {
           cases: {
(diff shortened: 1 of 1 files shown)


swoboda1337:
I also think it should be looking for tabs as well. I tested this change and it works. Without this if tabs are used then blocks are not recognized. If desired I can update the PR with this change as well:

```
     // First line of a Block Style
-    multiString: [[/^( +).+$/, "string", "@multiStringContinued.$1"]],
+    multiString: [[/^([ \t]+).+$/, "string", "@multiStringContinued.$1"]],

     // Further lines of a Block Style
     //   Workaround for indentation detection
     multiStringContinued: [
       [
-        /^( *).+$/,
+        /^([ \t]*).+$/,
         {
           cases: {
-            "$1~$S2 *": "string",
+            "$1~$S2[ \t]*": "string",
             "@default": { token: "@rematch", next: "@popall" },
           },
         },
```

jesserockz (approved):

PR #687, opened 2024-12-04
Add support for c++ syntax highlighting in lambdas

swoboda1337:
To simplify things only lambdas tagged with !lambda will be parsed as C++, automatically detecting the payload type in the dashboard is not feasible. IMO this is not a big deal and if someone wants the highlighting they will add the tag. 

Some people use the dashboard as their editor. It would be really nice to support embedded C++ syntax highlighting. PR is fairly simple with less than 50 new lines plus the C++ language definition taken from: https://github.com/microsoft/monaco-editor/blob/main/src/basic-languages/cpp/cpp.ts

Updated existing multi line block code to use the same regex as the new embedded ones. For consistency and fixes a small issue with a line of just spaces which would prematurely end the block. 

Everything has been tested and is working: 

image

Longer example:

image

Vs current dashboard:

image

- fixes https://github.com/esphome/feature-requests/issues/2012


clowrey:
Just wanted to say thanks again for getting this feature added!!! Just updated my dev today and it is glorious :D

PR #707, opened 2025-01-15
[feat] Add esptool logs to install web dialog

pzich:
This adds the output of the `esptool` logs to the installer dialog

https://github.com/user-attachments/assets/20da1bfe-d997-40ca-bce4-0415415a7ced


ximex:
@balloob could we get that merged?

src/install-web/install-web-dialog.ts
```diff
@@ -56,6 +61,23 @@ export class ESPHomeInstallWebDialog extends LitElement {
 
   private _platform?: ValueOf<typeof chipFamilyToPlatform>;
 
+  @property() private _showLogs = false;
+
+  connectedCallback() {
+    super.connectedCallback();
+
+    // this._console won't be defined until after the first paint
+    setTimeout(() => {
```
balloob, line 70: `this.updateComplete` is a promise that resolves once update is complete. There is also an `updated` callback.

src/install-web/install-web-dialog.ts
```diff
@@ -148,7 +180,7 @@ export class ESPHomeInstallWebDialog extends LitElement {
         <div class="icon">${icon}</div>
         ${label}
       </div>
-      ${showClose
+      ${showClose || true
```
balloob, line 183: dev?

pzich:
Yeah there's still some polish needed here. Hoping to take another look at this soon.

Gunni:
Can you remove the progress bar? Just entirely reuse the same log window "wireless" flashing uses?

Also, when flashing is done, start serial log immediately so we can catch the first boot after flashing.

src/install-web/install-web-dialog.ts
pzich: Whoops, yes, fixed.

src/install-web/install-web-dialog.ts
pzich: Aha, thanks! Switched to this

pzich:
> Can you remove the progress bar?

I think it still makes sense to have this as a "user friendly" / less technical option.

> Just entirely reuse the same log window "wireless" flashing uses?

I'm not sure if there's a good way to hand off the in progress serial log from the normal install dialog to the logs dialog. I'll try to see what's possible.

> Also, when flashing is done, start serial log immediately so we can catch the first boot after flashing.

I'm not sure if `esptool.js` supports this, but I'll also try to see what's possible here.

PR #710, opened 2025-01-23
improve esp32 schema

ximex:
- add schema definition for `flash_size` and `partitions`
- add missing type `string`
- add `esp32-c6` boards


ximex:
@jesserockz @balloob pls review and merge/add change requests

balloob (approved):

PR #738, opened 2025-04-21
use template strings

ximex:
use template strings for all string concatinations


ximex:
@balloob ready for review/merge

balloob (approved):
Thanks!

PR #740, opened 2025-04-30
Fix ColoredConsole performance issues

swoboda1337:
When doing `update all` the logs can eventually get minutes behind and logging can even stop before the final summary. In addition to `update all` being unusable `web.esphome.io` would crash with a high volume of logs. 

The old code was adding an empty span to the live page. That span was then updated multiple times causing the browser to constantly re-render. Significant savings came from building the full span first and only adding the line after the span was complete. 

Adding lines and updating the scroll position can take a long time. Instead of adding each line immediately, lines are now buffered. Processing is deferred until the current call stack is complete. This way if multiple lines are received at the same time those lines will be added together in one call saving a lot of time. 

Before this change (`update all` with three devices):
  - time spent in addLine: 157.656 secs
  - total logging time: 194.428 secs

After this change (`update all` with three devices):
  - time spent in processLines: 24.794 secs
  - total logging time: 97.277 secs
  
Testing was done using IDF, which has tons of files and on a fast build machine. Instead of being almost two minutes behind the logs now keep up with the build. 

Validating the config was also an interesting test. Before these changes it took almost 10 seconds to log the full config. After this change the full config is logged instantly.  

The dashboard log viewer and `web.esphome.io` can handle a significantly higher volume of logs now. The dashboard can now keep up with putty even after 100k lines of scroll back. Instead of crashing the web browser like it previously did `web.esphome.io` can now handle high volumes of logging with web serial. Previously the code was stuck reading serial + adding lines, the call stack never returned as logging was so slow that serial data backed up looping forever. Now all serial date is buffered, the call stack returns and only then are the new lines rendered.

### DIFF
--- a/src/util/console-color.ts
+++ b/src/util/console-color.ts
@@ -227,7 +227,6 @@ export const coloredConsoleStyles = `
     line-height: 1.45;
     border-radius: 3px;
     white-space: pre-wrap;
-    will-change: scroll-position;
     overflow-wrap: break-word;
     color: #ddd;
   }
--- a/src/util/console-color.ts
+++ b/src/util/console-color.ts
@@ -207,7 +207,7 @@ export class ColoredConsole {
   }
 
   addLine(line: string) {
-    // processing of lines is deferred for performance reasons
+    // Processing of lines is deferred for performance reasons
     if (this.state.lines.length == 0) {
       setTimeout(() => this.processLines(), 0);
     }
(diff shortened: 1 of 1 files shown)


swoboda1337:
@randybb I know you have seen this as well

swoboda1337:
This is done now and fully tested.

balloob (approved):
Nice!

PR #742, opened 2025-05-05
Increase serial buffer size

swoboda1337:
The default serial buffer size is 255 bytes which is only ~22 ms of logging data. With a decent amount of scroll back or some random delay data can get dropped. 

Increase this to a more reasonable amount for something like a browser, 8192 bytes is ~0.7 s of buffering.


balloob:
Would you be interested in porting your recent changes for the console and this PR to https://github.com/esphome/esp-web-tools too ?

swoboda1337:
> Would you be interested in porting your recent changes for the console and this PR to https://github.com/esphome/esp-web-tools too ?

Yeah I can do that. Code seems pretty similar.

PR #782, opened 2025-08-12
Add upload and empty config to the new device wizard

maximmaxim345:
Implements https://github.com/esphome/backlog/issues/7

Depends on https://github.com/esphome/esphome/pull/10203

# Overview

This PR add another step to the new device wizard where the user can now select how the configuration should be created:
- New Device Setup: continues the wizard as before
- Import from file: asks the user to upload an existing .yaml configuration file
- Empty configuration: creates a empty .yaml file for manual writing or pasting from https://devices.esphome.io/

Dropping a yaml file over this dialog is also supported.

# Screenshot

image

### DIFF
--- a/src/wizard/wizard-dialog.ts
+++ b/src/wizard/wizard-dialog.ts
@@ -710,6 +710,7 @@ export class ESPHomeWizardDialog extends LitElement {
 
   private async _handlePickBoardSubmit() {
     if (!this._board) return;
+    if(this._data.type !== "basic") return;
     this._data.board = this._board!;
 
     this._busy = true;
@@ -742,6 +743,7 @@ export class ESPHomeWizardDialog extends LitElement {
     this._error = undefined;
     let esploader: ESPLoader | undefined;
     let removeConfig = false;
+    if(this._data.type !== "basic") return;
     try {
       let port: SerialPort | undefined;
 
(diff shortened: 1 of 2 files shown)


maximmaxim345:
What would be really nice is to expand this wizard in the future and integrate the [Device Configuration Repo](https://devices.esphome.io/) into it somehow.

This would allow you to search for your device from the dashboard and import the configuration directly into ESPHome.

Custom templates (or perhaps user-defined custom repos?) could then also be added.

The `Empty Configuration` is just a stopgap to avoid having to click through the device options if you're going to overwrite it by pasting anyway.

randybb:
Yea, similar to https://uiflow2.m5stack.com/ - where you can select your device, then a template and go :)

Some people would argue that people who have many devices already have their own way (scripts, or custom ui), but why not bring this to all people? The UI can be customized - super simple enabled by default and an advanced one with bells and whistles.

src/wizard/wizard-dialog.ts
```diff
@@ -597,6 +708,47 @@ export class ESPHomeWizardDialog extends LitElement {
...
+    }
+  }
+
+  private async _handleConfigFileUpload() {
+    const input = this._configFileInput;
+    const file = input.files?.[0];
+    const response = await createConfiguration({
+      type: "upload",
+      name: file?.name.replace(/\.ya?ml$/, ""),
+      file_content: btoa(file ? await file.text() : ""),
```
jesserockz, line 744: Looks like btoa doesnt work on utf-8 files and just expects binary, which means the unicode symbols get mangled.

jesserockz (changes requested):

PR #810, opened 2025-09-22
Add clean all option

swoboda1337:
Recent platforms sometimes require platformio to be cleaned. Currently the only way to do that is by uninstalling and reinstalling the addon (or manually cleaning). Add an option to clean platformio and the full build dir.

- Requires https://github.com/esphome/esphome/pull/10876
- Resolves https://github.com/esphome/backlog/issues/61
- Resolves https://github.com/orgs/esphome/discussions/3217

image
image
image
image


swoboda1337:
> Maybe it should go as an option under kebab button as it is only a button for fixing stuff.

Could but then no one will see it and instead of fixing problems themselves they will open issues or come to discord.

swoboda1337:
Moved to just one option the nuclear one

PR #756, opened 2025-06-16
Integrate Home Assistant data table with enhanced UI features

brayStorm:
## Summary
- Integrate Home Assistant style data table component for device list view
- Add text-based online/offline status indicators with visual dots for better accessibility
- Implement comprehensive table view with sorting and filtering capabilities  
- Fix mDNS display to use configuration filename instead of device name
- Enhance card layout with improved grid responsiveness and theming
- Fix icon colors in dropdown menus for better visibility

## Key Features
- **HA Data Table Integration**: Implements Home Assistant's data table pattern for consistent UX across ESPHome and HA ecosystem
- **Dual View Modes**: Toggle between card view and table view with persistent user preferences
- **Advanced Sorting**: Sort by device name, IP address, or connection status with proper status ordering
- **Status Filtering**: Filter devices by online/offline status for easier management
- **Responsive Design**: Adaptive grid layout supporting 1-5 columns based on screen size
- **Enhanced Accessibility**: Text-based status indicators combined with visual dots

## Technical Implementation
- New `esphome-data-table` component following HA design patterns
- Device IP resolution API integration for accurate network information
- LocalStorage-based preference persistence for view settings
- Comprehensive CSS theming supporting both light and dark modes
- Proper mDNS address formatting using configuration filenames

## Test Plan
- [x] Verify HA-style data table renders correctly with proper theming
- [x] Test table sorting by name, IP address, and status with correct ordering
- [x] Confirm status filtering works for online/offline devices
- [x] Verify text status indicators display correctly in both card and table views
- [x] Confirm mDNS addresses show configuration filename (e.g., "device-name.local")
- [x] Check responsive grid layout across different screen sizes (1-5 columns)
- [x] Validate icon colors in dropdown menus are properly visible in all themes
- [x] Test preference persistence across browser sessions
- [x] Verify compatibility with existing device management workflows

PR #842, opened 2025-12-22
[Dashboard] Add online/offline status filter for devices

tsgoff:
## Description

This PR adds a status filter to the ESPHome Dashboard device list, allowing users to filter devices by their online/offline status.

### Changes Made

- Added three filter buttons to the device list: "All", "Online", and "Offline"
- Filter buttons appear between the search bar and the device grid
- Filter applies only to configured devices (importable/discovered devices are not affected)
- Active filter button is visually highlighted with the primary theme color
- Filter selection persists across page reloads using localStorage

### Implementation Details

- Added `StatusFilter` type: `"all" | "online" | "offline"`
- New `_statusFilter` state property to track the current filter selection
- Enhanced `_filter()` method to check device online status from `_onlineStatus`
- Filter state is saved to localStorage as `esphome-status-filter`
- On component initialization, the saved filter is restored from localStorage

### Use Case

Users with many ESPHome devices can now quickly:
- View only online devices to check active devices
- View only offline devices to identify devices that need attention
- Toggle between views without losing their search query

### Testing

- Built successfully with `script/build`
- Filter UI renders correctly above device grid
- Filter logic correctly shows/hides devices based on online status
- Filter selection persists after page reload
- Works alongside existing search functionality

### DIFF
--- a/src/devices/devices-list.ts
+++ b/src/devices/devices-list.ts
@@ -114,12 +114,16 @@ class ESPHomeDevicesList extends LitElement {
           @click=${() => this._handleStatusFilterChange("all")}
         ></mwc-button>
         <mwc-button
-          class="filter-button ${this._statusFilter === "online" ? "active" : ""}"
+          class="filter-button ${this._statusFilter === "online"
+            ? "active"
+            : ""}"
           label="Online"
           @click=${() => this._handleStatusFilterChange("online")}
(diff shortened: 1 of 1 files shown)


balloob:
We don't want to introduce a new filter bar right now. We rather have #756 land.

PR #902, opened 2026-04-25
Add 'Show entity state changes' toggle to OTA log viewer

bdraco:
## Context

The dashboard is slated to be replaced by a new system in the future. This PR is the minimal short-term change needed to fill the gap so users running the current dashboard have a way to opt out of entity-state log spam without dropping to a shell.

## Summary

ESPHome 2026.4.0 introduced `[S][sensor]` / `[S][text_sensor]` entity-state log lines that aren't suppressed by the existing `logger:` level/tag filters. The CLI ships `esphome logs --no-states` for this, but the dashboard had no UI escape hatch (issues [esphome/esphome#15987](https://github.com/esphome/esphome/issues/15987) and [esphome/esphome#15865](https://github.com/esphome/esphome/issues/15865)).

This adds a **"Show entity state changes"** checkbox to the bottom-left of the logs dialog. Toggling it sends `no_states` in the WebSocket spawn message (handled by the companion backend PR [esphome/esphome#15993](https://github.com/esphome/esphome/pull/15993)) and reconnects the log stream so the change takes effect immediately. The choice persists in `localStorage`.

The toggle is gated to OTA targets only — `--no-states` has no effect on serial logs (`run_miniterm` ignores it) and WebSerial logs run entirely in the browser.

Screenshot 2026-04-25 at 7 51 12 AM

Mobile

Screenshot 2026-04-25 at 8 45 16 AM

## Layout / refactor notes

- mwc-dialog's actions footer is hidden via the existing `hideActions` property; the entire bottom row (checkbox, Download Logs, secondaryAction slot, Close) is owned by `process-dialog` so the checkbox can sit at the dialog's left edge with the buttons flush right. Trying to do this via mwc-dialog's `primaryAction` / `secondaryAction` slots fights `mdc-dialog__actions { justify-content: flex-end }` and the `#actions ::slotted(*) { max-width: 100%; text-align: right }` rules — owning the row gives deterministic layout.
- `esphome-remote-process` gets a public `restart()` and tags each `streamLogs()` call with its `AbortController` so the stale connection's resolution can't fire `process-done` after the new one is started.
- `process-dialog` gets `show-states-toggle` / `.showStates` properties and emits `show-states-changed`; `logs-dialog` owns the persisted state and orchestrates the restart.

## Test plan

- [x] Open Logs on an OTA-reachable device — checkbox appears at left, buttons at right, single row.
- [x] Untick the checkbox — `[S][sensor]` lines stop streaming; backend log shows `--no-states` in the spawned argv.
- [x] Re-tick — state lines resume.
- [x] Close + reopen the dialog — checkbox state persists (`localStorage.esphome-logs-show-states`).
- [x] ESC key and scrim click still close the dialog.
- [x] Open Logs against a server serial port (`/dev/ttyXXX`) — checkbox is hidden.
- [x] WebSerial path ("Plug into this computer") is untouched (separate dialog).

### DIFF
--- a/src/components/process-dialog.ts
+++ b/src/components/process-dialog.ts
@@ -43,7 +43,7 @@ export class ESPHomeProcessDialog extends LitElement {
             @process-done=${this._handleProcessDone}
           ></esphome-remote-process>
 
-          <div class="action-row">
+          <div class="action-row" @click=${this._handleActionRowClick}>
             ${this.showStatesToggle
               ? html`
                   <mwc-formfield label="Show entity state changes">
@@ -74,6 +74,17 @@ export class ESPHomeProcessDialog extends LitElement {
 
   private _closeDialog() {
     this.shadowRoot?.querySelector("mwc-dialog")?.close();
+  }
+
+  // mwc-dialog's built-in dialogAction click delegation only walks closest()
+  // from mdcRoot's listener, which can't reach slotted buttons now that we
+  // render the action row in the content slot. Mirror the behaviour here so
+  // existing `dialogAction="close"` buttons keep closing the dialog.
+  private _handleActionRowClick(ev: MouseEvent) {
+    const target = ev.target as Element | null;
+    if (target?.closest("[dialogAction]")) {
+      this._closeDialog();
+    }
   }
 
   public restart() {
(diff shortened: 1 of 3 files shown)


balloob (approved):

bdraco:
thanks

PR #923, opened 2026-06-19
Add USB flash hand-off from ESPHome Device Builder via postMessage

bdraco:
## What

ESPHome Device Builder can build firmware but cannot flash over USB when it runs over plain http (the Home Assistant add-on), because Web Serial needs a secure context. This adds a mode where Device Builder opens web.esphome.io in a new tab and hands it the compiled firmware over postMessage, so the flashing runs here in a secure context.

When the page is opened with a nonce in the URL hash, it announces itself to the opener, accepts the firmware frame only from its opener window and only when the one time nonce matches (the opener origin is arbitrary, so there is no origin allowlist), converts the transferred image to the form install-web expects, and runs the existing connect and install flow. The firmware never touches a server.

## Post-flash reset (affects all installs)

This also replaces install-web's bare-RTS `resetSerialDevice` with a reset strategy that boots the app (RTC-watchdog for S2/S3/C2/C3, `UsbJtagSerialReset` for native-USB chips, classic EN-pulse with GPIO0 released for UART bridges), mirroring the device-builder-frontend web-serial reset. Native USB-Serial-JTAG chips previously stayed in download mode after writeFlash; this fixes the normal Connect & Install flow too, not just the hand-off. Verified on an ESP32-S3 over USB-Serial-JTAG and on a classic chip over a UART bridge.

## Notes

Companion change is in esphome/device-builder-frontend#904; tracked in esphome/backlog#151. Opening as a draft while the Device Builder side is finalized.

### DIFF
--- a/web.esphome.io/src/dashboard/ew-dashboard.ts
+++ b/web.esphome.io/src/dashboard/ew-dashboard.ts
@@ -112,15 +112,22 @@ class EWDashboard extends LitElement {
     `;
   }
 
-  protected firstUpdated(changedProps: PropertyValues): void {
-    super.firstUpdated(changedProps);
-    // Only enter flash mode for a real hand-off (opened by Device Builder). A
-    // stale #nonce bookmark without an opener falls through to the dashboard.
+  connectedCallback(): void {
+    super.connectedCallback();
+    // Decide flash mode before the first render to avoid a one-frame flash of
+    // the normal dashboard. Only a real hand-off (opened by Device Builder) qualifies;
+    // a stale #nonce bookmark without an opener falls through to the dashboard.
     if (
       window.opener &&
       new URLSearchParams(location.hash.slice(1)).get("nonce")
     ) {
       this.flashMode = true;
+    }
+  }
+
+  protected firstUpdated(changedProps: PropertyValues): void {
+    super.firstUpdated(changedProps);
+    if (this.flashMode) {
       return;
     }
     const params = new URLSearchParams(location.search);
--- a/web.esphome.io/src/dashboard/ew-dashboard.ts
+++ b/web.esphome.io/src/dashboard/ew-dashboard.ts
@@ -4,6 +4,7 @@ import "@material/mwc-button";
 import "./unsupported-card";
 import "./esp-connect-card";
 import "./pico-connect-card";
+import "./ew-web-flash";
 import { supportsWebSerial } from "../../../src/const";
 import { ARROW, ARROW_RIGHT } from "./arrow";
 import { esphomeSvgStyles } from "../../../src/styles";
@@ -11,8 +12,21 @@ import { esphomeSvgStyles } from "../../../src/styles";
 @customElement("ew-dashboard")
 class EWDashboard extends LitElement {
   @state() private pico = false;
+  // Set when opened by ESPHome Device Builder to receive firmware over
+  // postMessage and flash it over USB (the add-on is plain http, so it can't run
+  // Web Serial itself); the nonce in the hash is the hand-off marker.
+  @state() private flashMode = false;
 
   protected render() {
+    if (this.flashMode) {
+      return html`
+        <div class="container">
+          ${!supportsWebSerial
+            ? html`<ew-unsupported-card></ew-unsupported-card>`
+            : html`<ew-web-flash></ew-web-flash>`}
+        </div>
+      `;
+    }
     return html`
       <div class="container">
         ${!supportsWebSerial
@@ -100,6 +114,10 @@ class EWDashboard extends LitElement {
 
   protected firstUpdated(changedProps: PropertyValues): void {
     super.firstUpdated(changedProps);
+    if (new URLSearchParams(location.hash.slice(1)).get("nonce")) {
+      this.flashMode = true;
+      return;
+    }
     const params = new URLSearchParams(location.search);
 
     for (const key of params.keys()) {
--- a/web.esphome.io/src/dashboard/ew-dashboard.ts
+++ b/web.esphome.io/src/dashboard/ew-dashboard.ts
@@ -114,7 +114,12 @@ class EWDashboard extends LitElement {
 
   protected firstUpdated(changedProps: PropertyValues): void {
     super.firstUpdated(changedProps);
-    if (new URLSearchParams(location.hash.slice(1)).get("nonce")) {
+    // Only enter flash mode for a real hand-off (opened by Device Builder). A
+    // stale #nonce bookmark without an opener falls through to the dashboard.
+    if (
+      window.opener &&
+      new URLSearchParams(location.hash.slice(1)).get("nonce")
+    ) {
       this.flashMode = true;
       return;
     }
--- a/web.esphome.io/src/dashboard/ew-web-flash.ts
+++ b/web.esphome.io/src/dashboard/ew-web-flash.ts
@@ -112,6 +112,7 @@ class EWWebFlash extends LitElement {
       return this._port
         ? html`<ew-esp-device-card
             .port=${this._port}
+            .name=${this._deviceName}
             @close=${this._onCardClose}
           ></ew-esp-device-card>`
         : html`<esphome-card status="DONE">
--- a/web.esphome.io/src/dashboard/ew-web-flash.ts
+++ b/web.esphome.io/src/dashboard/ew-web-flash.ts
@@ -70,8 +70,9 @@ class EWWebFlash extends LitElement {
 
   private _nonce = "";
   private _opener: Window | null = null;
-  // Stays '*' until the first valid inbound frame reveals the opener origin; no
-  // outbound frame carries the nonce, so the pre-handoff broadcast leaks nothing.
+  // Seeded from the 'origin' hash param when the opener passes it, otherwise '*'
+  // until the first valid inbound frame reveals the opener origin. No outbound
+  // frame carries the nonce, so even the '*' broadcast leaks nothing.
   private _targetOrigin = "*";
   private _files: FileToFlash[] | null = null;
   private _erase = true;
--- a/web.esphome.io/src/dashboard/ew-web-flash.ts
+++ b/web.esphome.io/src/dashboard/ew-web-flash.ts
@@ -247,59 +247,66 @@ class EWWebFlash extends LitElement {
     // (logs are a separate fresh connect), mirroring the device-builder flasher.
     // "Erasing" until the first write progress arrives, then "Installing".
     let writing = false;
-    await openInstallWebDialog({
-      erase: this._erase,
-      filesCallback: async () => files,
-      // Mirror progress and the granular phase so the dashboard's install view
-      // tracks this tab instead of showing a stale phase.
-      onProgress: (pct) => {
-        this._post({ type: MSG_PROGRESS, pct });
-        if (!writing) {
-          writing = true;
-          this._post({
-            type: MSG_STATE,
-            state: "installing",
-            detail: "Installing over USB…",
-          });
-        }
-      },
-      onStateChange: (state, error) => {
-        if (state === "done") {
-          this._post({
-            type: MSG_STATE,
-            state: error ? "error" : "done",
-            detail: error ?? "",
-          });
-          return;
-        }
-        // Map install-web's phases to the mirror; onProgress flips to
-        // "Installing" once the first write lands.
-        if (state === "installing") {
-          this._post({
-            type: MSG_STATE,
-            state: "installing",
-            detail: "Erasing…",
-          });
-        } else if (state === "prepare_installation") {
-          this._post({
-            type: MSG_STATE,
-            state: "installing",
-            detail: "Preparing…",
-          });
-        } else {
-          this._post({
-            type: MSG_STATE,
-            state: "connecting",
-            detail: "Connecting to device…",
-          });
-        }
-      },
-      onClose: (success) => {
-        this._busy = false;
-        // Failed: stay on the card so the user can retry. Succeeded: done.
-        this._status = success ? "done" : "ready";
-      },
-    });
+    try {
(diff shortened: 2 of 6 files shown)


web.esphome.io/src/dashboard/ew-dashboard.ts
```diff
@@ -100,6 +114,10 @@ class EWDashboard extends LitElement {
 
   protected firstUpdated(changedProps: PropertyValues): void {
     super.firstUpdated(changedProps);
+    if (new URLSearchParams(location.hash.slice(1)).get("nonce")) {
+      this.flashMode = true;
+      return;
+    }
```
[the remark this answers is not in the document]
bdraco: Fixed in 94489d4 — flash mode is now gated on window.opener, so a stale #nonce link falls through to the normal dashboard.

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,225 @@
...
+  private _stopReady(): void {
+    if (this._readyTimer !== undefined) {
+      clearInterval(this._readyTimer);
+      this._readyTimer = undefined;
+    }
+  }
+
+  private _post(message: object): void {
+    this._opener?.postMessage(message, this._targetOrigin);
+  }
```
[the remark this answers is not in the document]
bdraco: Fixed in 94489d4 — _post now wraps postMessage in try/catch and falls back to '*' so a malformed origin can't wedge the ready interval.

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,225 @@
...
+    );
+    // Port picker dismissed or failed before the dialog opened: re-enable so the
+    // user can try again (install-web's own no-port dialog also offers a retry).
+    if (!opened) this._busy = false;
+  };
+
+  private _fail(detail: string): void {
+    this._status = "error";
+    this._detail = detail;
+  }
```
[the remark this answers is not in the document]
bdraco: Fixed in 94489d4 — _fail now calls _stopReady(), so the ready announcements stop once the tab is in an error state.

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,225 @@
...
+// doesn't blow the argument limit of String.fromCharCode.
+function bufferToBinaryString(buffer: ArrayBuffer): string {
+  const bytes = new Uint8Array(buffer);
+  const CHUNK = 0x8000;
+  let result = "";
+  for (let i = 0; i < bytes.length; i += CHUNK) {
+    result += String.fromCharCode(...bytes.subarray(i, i + CHUNK));
+  }
+  return result;
+}
```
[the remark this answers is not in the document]
bdraco: Addressed in 94489d4 — replaced the chunked loop with new TextDecoder('latin1').decode(buffer); one native call, byte-identical to readAsBinaryString, no concat.

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,225 @@
...
+    try {
+      this._files = (data.parts as FirmwarePart[]).map((part) => ({
+        data: bufferToBinaryString(part.data),
+        address: part.address,
+      }));
+    } catch {
+      this._fail("Could not read the firmware payload.");
+      return;
+    }
+    this._erase = data.erase ?? true;
```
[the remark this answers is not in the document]
bdraco: Fixed in 94489d4 — erase is now coerced with data.erase !== false (only an explicit boolean false disables erase).

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,225 @@
...
+    }
+    if (this._targetOrigin === "*" && ev.origin && ev.origin !== "null") {
+      this._targetOrigin = ev.origin;
+    }
+    this._stopReady();
+    try {
+      this._files = (data.parts as FirmwarePart[]).map((part) => ({
+        data: bufferToBinaryString(part.data),
+        address: part.address,
+      }));
```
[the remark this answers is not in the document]
bdraco: Fixed in 94489d4 — added isFlashParts() validating each part's address (non-negative integer) and data (ArrayBuffer) before conversion, mirroring the device-builder flasher.

esphbot (changes requested):
Blocking issues found — see the review comment above.

bdraco:
Thanks Kōan — this pass ran on `10d5a8e` (the initial commit); all of these were addressed in `94489d4` and the branch is now at `dd866a2`:

- **Gate flash mode on `window.opener`** (#1) — done; a stale `#nonce` link falls through to the dashboard.
- **postMessage throws on a malformed origin** (#2) — `_post` wraps it in try/catch and falls back to `"*"`.
- **`_fail` leaves the READY interval running** (#3) — `_fail` now calls `_stopReady()`.
- **Per-part validation + erase coercion** (#4) — added `isFlashParts()` (non-negative integer `address`, `data instanceof ArrayBuffer`) and `data.erase !== false`.

On `PROTOCOL_VERSION`: it's advertised but intentionally not enforced while there's a single version; happy to add a check when v2 lands. The string-concat note I left as `String.fromCharCode` (a later `TextDecoder("latin1")` attempt corrupted firmware, since that label is windows-1252 per the Encoding spec).

Later commits also fixed the real on-hardware issues found in testing: the post-flash reset (native USB-Serial-JTAG was landing in download mode), the install-web port reopen that re-reset the chip into a boot loop, and treating the reset's re-enumeration as "Device disconnected".

esphbot (approved):
No blocking issues found.

bdraco:
Thanks Kōan — both cosmetic suggestions applied:

- Flash mode is now decided in `connectedCallback` (before the first render), so a real hand-off no longer flickers the normal dashboard for a frame; `firstUpdated` early-returns in flash mode.
- Updated the `_targetOrigin` comment to note it's seeded from the `origin` hash param when present, otherwise `'*'` (keeping the no-nonce-outbound reasoning).

Also note the `_post` origin fallback was just hardened (66049d6) to log rather than swallow non-origin errors and to guard the fallback call, matching the same fix on the device-builder flasher (#1609).

esphbot (changes requested):
Blocking issues found — see the review comment above.

bdraco:
Thanks Kōan:

- **Terminal state on timeout (suggestion):** fixed in 8edc7c1 — when the 10s announce window elapses with no firmware, the card now flips to an error ("Didn't receive firmware. Return to ESPHome Device Builder and try the USB install again.") instead of spinning forever.

- **hardResetChip on the shared install path (important):** intentional, not scope creep. The bare-RTS `resetSerialDevice` left native USB-Serial-JTAG chips (S3/C3) in download mode after writeFlash on the *normal* Connect & Install too, so scoping the fix to only the hand-off would leave the main web flasher broken for those chips. The new strategy is the same one `esphome/device-builder-frontend`'s `web-serial.ts` already uses in production (RTC-WDT for S2/S3/C2/C3, `UsbJtagSerialReset` for native USB, `classicHardReset` otherwise), and `classicHardReset` additionally releases GPIO0 (DTR low) which the old bare-RTS reset never did, so it's an improvement for UART-bridge ESP32/ESP8266, not just native USB. Verified on an S3 over USB-Serial-JTAG; classic-bridge (CP2102/CH340) verification on the normal flow is worth a maintainer pass before merge. I'll note this in the PR description so the scope is explicit.

- **PROTOCOL_VERSION:** now exchanged both directions (added `FirmwareMessage.version`), but enforcement is intentionally deferred until a breaking change needs it; the additive-compatibility contract and the version-as-lever rule are documented in esphome/device-builder#1611.

bdraco:
Confirmed: the shared hardResetChip reset is verified on both an ESP32-S3 (native USB-Serial-JTAG) and a classic chip over a UART bridge, so the normal Connect & Install path is covered, not just the hand-off.

esphbot (changes requested):
Blocking issues found — see the review comment above.

bdraco:
Thanks Kōan — fixed in f044250: the post-flash `hardResetChip` is now wrapped in try/catch, logging the error but still setting `_state = "done"`, since by `installed = true` the write is committed. A reset that throws (e.g. `classicHardReset` / `UsbJtagSerialReset.reset()`) no longer reports a successful flash as failed.

On the shared-reset gate: classic ESP32/ESP8266 over a UART bridge has now been verified on the normal Connect & Install flow (in addition to the S3 over native USB-Serial-JTAG), so the main flasher path is covered.

esphbot (approved):
No blocking issues found.

bdraco:
Thanks Kōan — applied in the latest commit: prepare_installation now mirrors as "Preparing…" instead of "Connecting to device…". Checklist's all green; remaining gate is the companion #904 + the production FLASHER_URL flip, not code.

esphbot (approved):
No blocking issues found.

bdraco:
Acknowledged — both are low-severity and non-blocking; leaving as-is for now (the watchdogReset early-write guard is a cheap follow-up if we ever see it in practice).

bdraco:
Update: went ahead and fixed both in 3c146da. (1) watchdogReset only tolerates the final re-lock write racing the reset; a transport error before the WDT is armed now returns false so hardResetChip falls through to UsbJtagSerialReset (no silent download-mode). (2) ew-web-flash._install wraps openInstallWebDialog so a throw before onClose re-enables the button instead of leaving it disabled.

esphbot (approved):
No blocking issues found.

bdraco:
Fixed in 8cc5e27: _fail() now posts { type: 'esphome-web-flash:state', state: 'error', detail } to the opener, so a post-handoff receiver failure (malformed payload / conversion throw) surfaces on the dashboard instead of hanging its install view. It's a no-op when there's no opener (the open-directly error case). The watchdogReset fall-through you noted was already in 3c146da.

esphbot (approved):
No blocking issues found.

bdraco:
Done in the latest commit: wrapped the config0 expression in `>>> 0` so it reads as an unsigned u32 (0xD0000102) instead of a negative signed int, with a comment naming the fields (enable / stage0=reset-system / clock prescaler / timeout-stage). Same bit pattern esptool-js packs, just no longer a maintenance trap.

esphbot (approved):
No blocking issues found.

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,354 @@
...
+  connectedCallback(): void {
+    super.connectedCallback();
+    const params = new URLSearchParams(location.hash.slice(1));
+    this._nonce = params.get("nonce") ?? "";
+    this._targetOrigin = params.get("origin") || "*";
+    this._opener = window.opener;
+    if (!this._opener || !this._nonce) {
+      this._status = "error";
+      this._detail =
+        "Open this page from ESPHome Device Builder to flash a device over USB.";
```
balloob, line 90: I don't think we have to aim for a user understandable error here. A user won't know how to open this page from EDB either?

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,354 @@
...
+  private _opener: Window | null = null;
+  // Seeded from the 'origin' hash param when the opener passes it, otherwise '*'
+  // until the first valid inbound frame reveals the opener origin. No outbound
+  // frame carries the nonce, so even the '*' broadcast leaks nothing.
+  private _targetOrigin = "*";
+  private _files: FileToFlash[] | null = null;
+  private _erase = true;
+  private _readyTimer?: number;
+
+  connectedCallback(): void {
```
balloob, line 81: technically, `firstUpdated` is the correct LitElement way, but in this case we'll never connect more than once, so it's fine.

web.esphome.io/src/dashboard/ew-web-flash.ts
balloob: Guess it cna't hurt.

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,354 @@
...
+    if (this._status === "done") {
+      return html`<esphome-card status="DONE">
+        <div class="card-header">
+          ${this._deviceName
+            ? html`Installed ${this._deviceName}`
+            : "Install over USB"}
+        </div>
+        <div class="card-content">
+          Installed and rebooting. You can close this tab, or
+          <a href="${location.pathname}${location.search}">connect</a> to view
```
balloob, line 129: how is this page loaded? Should we open this on a new window or target _top ?

bdraco: Its already in a fresh tab so its refreshing into a clean state to reconnect

web.esphome.io/src/dashboard/ew-web-flash.ts
```diff
@@ -0,0 +1,354 @@
...
+    preloadInstallWebDialog();
+    // Re-announce until firmware arrives: a single 'ready' can race the opener
+    // attaching its listener after window.open().
+    this._sendReady();
+    let waited = 0;
+    this._readyTimer = window.setInterval(() => {
+      waited += 500;
+      if (this._files) {
+        this._stopReady();
+        return;
```
balloob: This can't happen, because we call `stopReady()` before we set files.

bdraco: agree its dead

bdraco: 14684ce34865b8974ff8a5b16f4abdd5abbb39ac

balloob (approved):

bdraco:
thanks

